PR #1, opened 2024-03-14
Subgraph implementation

Tanya-atatakai:
This was autogenerated based on the Depository contract.

Currently used only to get Create\Close product events like this:

```
query GetCreateProducts {
    createProducts(first: 1000) {
      productId
      token
      priceLP
      supply
      vesting
    }
  }
```

**note:** there's a limit on query (maximum is 1000) currently we have ~100-150 events for each type, in future we'll need to add a counter that stores total number of events and request events in portions using `skip` or `where` filtering: https://thegraph.com/docs/en/querying/graphql-api/#pagination

If you wish, you can check the subgraph yourself, to do that you need:
1) connect with your wallet here https://thegraph.com/studio
2) create a subgraph with any name
3) use step by step all in AUTH & DEPLOY part on the right
4) use playground for querying 
image


angrybayblade (approved):
Approving, but you don't need to upload the build and generated folders

PR #2, opened 2024-07-23
Emissions to devs

Tanya-atatakai:
1) Added Tokenomics and Dispenser contracts
2) Supported two different EpochSettled events
3) Added Epoch and DevIncentive entities so we can easily get needed data

Note: almost everything was auto-generated, I left comments of what can be reviewed

image


abis/Tokenomics.json
```diff
@@ -0,0 +1,1222 @@
...
+        "indexed": false,
+        "internalType": "uint256",
+        "name": "totalStakingIncentive",
+        "type": "uint256"
+      }
+    ],
+    "name": "EpochSettled",
+    "type": "event"
+  },
+  {
```
Tanya-atatakai, line 260: This was added manually to the latest abi, to support old EpochSettled event when there were less parameters

schema.graphql
```diff
@@ -75,3 +75,233 @@ type TreasuryUpdated @entity(immutable: true) {
...
+
+type DispenserTreasuryUpdated @entity(immutable: true) {
+  id: Bytes!
+  treasury: Bytes! # address
+  blockNumber: BigInt!
+  blockTimestamp: BigInt!
+  transactionHash: Bytes!
+}
+
+type Epoch @entity {
```
Tanya-atatakai, line 292: this and the DevIncentive entities were added manually

src/dispenser.ts
```diff
@@ -0,0 +1,113 @@
...
+  entity.reward = event.params.reward;
+  entity.topUp = event.params.topUp;
+
+  entity.blockNumber = event.block.number;
+  entity.blockTimestamp = event.block.timestamp;
+  entity.transactionHash = event.transaction.hash;
+
+  entity.save();
+
+  // Update Epoch entity with devIncentive
```
Tanya-atatakai, line 31: here and until the end of the function is the logic where I search for the epoch during which this event occurred and write needed data to the Epoch entity. you can see that I only update devIncentivesTotalTopUp in the epoch, but not devIncentives, but they will be automatically included in the epochs thanks to the [derivedFrom](https://thegraph.com/docs/en/developing/creating-a-subgraph/#reverse-lookups) relationship

src/tokenomics.ts
```diff
@@ -0,0 +1,406 @@
...
+  entity.returnedStakingIncentive = event.params.returnedStakingIncentive;
+  entity.totalStakingIncentive = event.params.totalStakingIncentive;
+
+  entity.blockNumber = event.block.number;
+  entity.blockTimestamp = event.block.timestamp;
+  entity.transactionHash = event.transaction.hash;
+
+  entity.save();
+
+  // Epoch entity
```
Tanya-atatakai, line 167: here I manually write the Epoch entity

src/tokenomics.ts
```diff
@@ -0,0 +1,406 @@
...
+  entity.accountRewards = event.params.accountRewards;
+  entity.accountTopUps = event.params.accountTopUps;
+
+  entity.blockNumber = event.block.number;
+  entity.blockTimestamp = event.block.timestamp;
+  entity.transactionHash = event.transaction.hash;
+
+  entity.save();
+
+  // Epoch entity
```
Tanya-atatakai, line 202: here I manually write the Epoch entity inside the old EpochSettled handler

truemiller (approved):

schema.graphql
```diff
@@ -75,3 +75,233 @@ type TreasuryUpdated @entity(immutable: true) {
...
+}
+
+type TokenomicsParametersUpdateRequested @entity(immutable: true) {
+  id: Bytes!
+  epochNumber: BigInt! # uint256
+  devsPerCapital: BigInt! # uint256
+  codePerDev: BigInt! # uint256
+  epsilonRate: BigInt! # uint256
+  epochLen: BigInt! # uint256
+  veOLASThreshold: BigInt! # uint256
```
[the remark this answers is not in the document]
Tanya-atatakai: the schema is generated automatically based on abis. I see only `blockNumber`, `blockTimestamp` and `transactionHash` don't have without such comments, probably because they are not event params but just stored as historical data via handlers (which is also auto generated :D)

PR #11, opened 2025-07-09
Feature/olas-holder-multichain

axatbhardwaj:
This PR enables the support of olas token holders on multiple chains

currently has support for 
- ETH
- Base
- Gnosis


src/olas.ts
```diff
@@ -1,20 +1,75 @@
...
+  let holderId = tokenAddress.toHex() + "-" + holderAddress.toHex();
+  let holder = TokenHolder.load(holderId);
+  let token = getOrCreateToken(tokenAddress);
+
+  if (holder == null) {
+    holder = new TokenHolder(holderId);
+    holder.token = tokenAddress;
+    holder.address = holderAddress;
+    holder.balance = BIGINT_ZERO;
+  }
```
Tanya-atatakai, line 48: could also have getOrCreateHolder function, wdyt?

schema.graphql
```diff
@@ -450,12 +450,20 @@ type BondClaim @entity(immutable: false) {
   blockTimestamp: BigInt!
 }
 
-type Token @entity {
+type Token @entity(immutable: false) {
   id: Bytes!
   balance: BigInt!
+  holderCount: BigInt
```
Tanya-atatakai, line 456:
```suggestion
  holderCount: BigInt!
```
we anyway set to 0 on token creation

Tanya-atatakai: also, could be just Int!

src/olas.ts
```diff
@@ -1,20 +1,75 @@
...
+  holder.balance = newBalance;
+
+  if (oldBalance.equals(BIGINT_ZERO) && newBalance.gt(BIGINT_ZERO)) {
+    token.holderCount = token.holderCount!.plus(BigInt.fromI32(1));
+  } else if (oldBalance.gt(BIGINT_ZERO) && newBalance.equals(BIGINT_ZERO)) {
+    token.holderCount = token.holderCount!.minus(BigInt.fromI32(1));
+    store.remove("TokenHolder", holderId);
+  } else {
+    holder.save();
+  }
```
Tanya-atatakai, line 67: will we ever be here? I think we should do `holder.save();` right after we do `holder.balance = newBalance;`

src/olas.ts
```diff
@@ -42,6 +97,22 @@ export function handleTransfer(event: Transfer): void {
...
+    token.balance = token.balance.plus(event.params.amount);
+  }
+  if (event.params.to.equals(ZERO_ADDRESS)) {
+    // Burn
+    token.balance = token.balance.minus(event.params.amount);
+  }
+  token.save();
+
+  updateTokenHolder(event.address, event.params.from, event.params.amount, true);
+  updateTokenHolder(event.address, event.params.to, event.params.amount, false);
```
Tanya-atatakai, line 114: shouldn't we check the sender instead of hardcoding true and false? or can you elaborate what this means?

src/olas.ts
```diff
@@ -1,20 +1,75 @@
...
   Address.fromString("0x52a043bcebdb2f939baef2e8b6f01652290eab3f"),
 ];
 
+function getOrCreateToken(tokenAddress: Address): Token {
+  let token = Token.load(tokenAddress);
+
+  if (token == null) {
+    token = new Token(tokenAddress);
+    let contract = OLAS.bind(tokenAddress);
+    token.balance = contract.totalSupply();
```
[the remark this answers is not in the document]
axatbhardwaj: yes it makes sense i'll set it to zero here

src/olas.ts
axatbhardwaj: 1. updateTokenHolder(event.address, event.params.from, event.params.amount, true);
    - Here, we are updating the balance for the from address, which is the account **sending** the tokens.
    - We pass true to tell the function, "Yes, this address is the sender," so it correctly **decreases** their balance.
2. updateTokenHolder(event.address, event.params.to, event.params.amount, false);
    - In the second call, we are updating the to address, which is the account **receiving** the tokens.
    - We pass false to tell the function, "No, this address is not the sender (it's the receiver)," so it correctly **increases** their balance.
    
    
   Either way this is not a good patter , i'll refactor this a little bit to stream line the approach

Tanya-atatakai (approved):

axatbhardwaj:
@Tanya-atatakai ,

Now it contains all the chains except solana

mariapiamo (approved):

PR #15, opened 2025-07-17
Feature/daa-subgraphs

axatbhardwaj:
This is a Restructure for Repo enabling DAA support

- The changes are huge in terms of num of lines but 60% of them are from ABIs
- added new .yaml files for various chains 
- it also include dynamic source template for newly created services

- there are only a few at the moment i'm planning to have more once this PR is reviewed and i've tested the existing ones


scripts/contract-creation-block-finder.js
```diff
@@ -0,0 +1,124 @@
+const { ethers } = require("ethers");
```
axatbhardwaj, line 1: @Tanya-atatakai ,

This is a script to find the block to find the startblock

PR #17, opened 2025-07-18
Feature/Daa-subgraphs-with-agent-ids

axatbhardwaj:
### This PR introduces the service registry subgraphs and fixes the build process for all subgraphs.
- Added service registry subgraphs for Ethereum, Gnosis, Optimism, Arbitrum, Base, Celo, Mode, and Polygon.
- Updated package.json with build scripts for the new subgraphs.
- Fixed incorrect paths in all subgraph.yaml files, resolving build failures.
- Updated the deploy-studio.js script to support deploying all subgraphs and improved the network selection UX.


subgraphs/service-registry/service-registry-arbitrum/subgraph.arbitrum.yaml
```diff
@@ -0,0 +1,58 @@
...
+dataSources:
+  # This is the STATIC data source. It listens to the main ServiceRegistryL2 contract
+  # to discover when new services are created.
+  - kind: ethereum/contract
+    name: ServiceRegistryL2
+    network: arbitrum-one
+    source:
+      address: "0xE3607b00E75f6405248323A9417ff6b39B244b50"
+      abi: ServiceRegistryL2
+      startBlock: 24205712
```
Tanya-atatakai, line 15: this block is too old, it'l index longer. let's use this start block https://arbiscan.io/tx/0x44ebcd93fe1c9dd61b73215ce8fafc0f0eb501d83cc2ad95ab186fd445850717

subgraphs/service-registry/service-registry-eth/subgraph.yaml
```diff
@@ -0,0 +1,58 @@
...
+    mapping:
+      kind: ethereum/events
+      apiVersion: 0.0.7
+      language: wasm/assemblyscript
+      entities:
+        - Service
+      abis:
+        # ABI for the main registry contract.
+        - name: ServiceRegistryL2
+          file: ../../../abis/ServiceRegistryL2.json
```
Tanya-atatakai, line 25: it has a little different ABI, on ethereum the contract is called ServiceRegistry. can we handle it separately? not sure if the events you track will be different though, but just in case

Tanya-atatakai (commented):
tiny suggestions

shared/service-registry/mappers/mapping.ts
```diff
@@ -0,0 +1,130 @@
...
+    service.multisig = multisig.id;
+
+    multisig.save();
+    service.save();
+
+    GnosisSafeTemplate.create(event.params.multisig);
+  }
+}
+
+export function handleRegisterInstance(event: RegisterInstance): void {
```
Tanya-atatakai, line 45: as I remember, Pia said that `RegisterInstance` happens before `CreateMultisigWithAgents`, but it seems we expect it vice versa and create the service entity in CreateMultisigWithAgents. is that correct?

shared/service-registry/mappers/mapping.ts
```diff
@@ -0,0 +1,130 @@
...
+import { GnosisSafe as GnosisSafeTemplate } from "../generated/templates";
+
+const ONE_DAY = BigInt.fromI32(86400);
+
+export function handleCreateMultisig(event: CreateMultisigWithAgents): void {
+  let multisigAddress = event.params.multisig;
+  let multisig = Multisig.load(multisigAddress);
+
+  if (multisig == null) {
+    let serviceId = event.params.serviceId.toString();
```
Tanya-atatakai, line 23: a service may exist when we change multisig afaik, can you clarify with Pia please? afaik the flow is:
1) service created and it has an id
2) agents are registered (agent Ids are assigned to the service)
3) multisig is created and assigned to the service
4) service is deployed

2-4 can repeat for the same service with different agent IDs and multisigs

rajat2502 (approved):

PR #20, opened 2025-07-25
Feature daa subgraph active multisig

axatbhardwaj:
This is the Final version of the subgraph :

- it returns all the 4 metrics successfully 
1. count of distinct active multisig per day per agent
2. count of district active multisig per day (across all the agents)
3. total txns per agent id
4. total txns per chain

### DIFF
--- a/shared/service-registry/mappers/mapping.ts
+++ b/shared/service-registry/mappers/mapping.ts
@@ -42,7 +42,7 @@ function getDailyUniqueAgentsId(event: ethereum.Event): string {
 
 function getDailyAgentPerformanceId(
   event: ethereum.Event,
-  agentId: BigInt
+  agentId: i32
 ): string {
   const dayTimestamp = getDayTimestamp(event);
   return "day-"
@@ -56,12 +56,17 @@ function updateDailyAgentPerformance(
   multisig: Multisig
 ): void {
   const dayTimestamp = getDayTimestamp(event);
+
+  // Process each agent associated with this multisig
   for (let i = 0; i < multisig.agentIds.length; i++) {
     const agentId = multisig.agentIds[i];
-    const id = getDailyAgentPerformanceId(event, BigInt.fromI32(agentId));
+
+    // Generate unique ID for this agent's daily performance
+    const id = getDailyAgentPerformanceId(event, agentId);
     let entity = DailyAgentPerformance.load(id);
 
     if (entity == null) {
+      // Create new daily performance entity for this agent
       entity = new DailyAgentPerformance(id);
       entity.agentId = agentId;
       entity.dayTimestamp = dayTimestamp;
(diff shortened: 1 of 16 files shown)


shared/service-registry/schema.graphql
```diff
@@ -19,17 +19,36 @@ type DailyServiceActivity @entity(immutable: true) {
...
+type DailyActiveMultisigs @entity(immutable: false) {
+  id: ID!
+  dayTimestamp: BigInt!
+  activeMultisigs: [Bytes!]!
+  count: Int!
 }
 
+type Agent @entity(immutable: false) {
+  id: ID! # agentId
+  totalTxCount: BigInt!
```
Tanya-atatakai, line 47: please let's ensure the fields that mean the same are called the same across different entitites
```suggestion
  txCount: Int!
```

shared/service-registry/mappers/mapping.ts
```diff
@@ -117,6 +159,45 @@ function updateDailyActivity(
...
+    dailyEntity.dayTimestamp = dayTimestamp;
+    dailyEntity.activeMultisigs = [];
+    dailyEntity.count = 0;
+  }
+
+  if (!dailyEntity.activeMultisigs.includes(multisig.id)) {
+    let multisigs = dailyEntity.activeMultisigs;
+    multisigs.push(multisig.id);
+    dailyEntity.activeMultisigs = multisigs;
+    dailyEntity.count = dailyEntity.activeMultisigs.length;
```
Tanya-atatakai, line 181: please could you remind me again, why it's needed? in which query we can use that?

shared/service-registry/schema.graphql
axatbhardwaj: we call it totalTxCount because it is the total from inception but in case of 

```
DailyAgentPerformance @entity(immutable: false) {
  id: ID!
  dayTimestamp: BigInt!
  agentId: Int!
  txCount: Int!
  activeMultisigs: [Bytes!]!
  uniqueActiveMultisigCount: Int!
}
```

txCount represents txcount on the sepcific day

shared/service-registry/mappers/mapping.ts
axatbhardwaj: to track the total number of distinct active multisigs per day across all agents

shared/service-registry/mappers/mapping.ts
```diff
@@ -12,151 +12,116 @@ import {
...
 }
 
 export function handleRegisterInstance(event: RegisterInstance): void {
-  let service = Service.load(event.params.serviceId.toString());
-  if (service != null) {
-    service.agentIds = [event.params.agentId.toI32()];
-    service.save();
-  }
+  let service = getOrCreateService(event.params.serviceId);
+  service.agentIds = [event.params.agentId.toI32()];
```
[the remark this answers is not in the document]
rajat2502: hmm, i tried that on my subgraph, unfortunately the totalTxns with this change didn't match our dune queries :(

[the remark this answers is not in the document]
axatbhardwaj: Yes you are right @rajat2502 

but the problem was with dune queries and pi fixed those since then i've implemented the changes and pushed here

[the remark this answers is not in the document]
rajat2502: i'm afraid it would still be different, anyways, we can try that on the studio

Tanya-atatakai (approved):

PR #21, opened 2025-07-28
Agent registration tracking

axatbhardwaj:
- Added AgentRegistration entity to track agent registration timestamps
- Updated handleCreateMultisig to use "most recent agent before deployment" logic matching SQL
- Each multisig now associates with exactly one agent instead of all service agents


shared/service-registry/mappers/mapping.ts
```diff
@@ -125,11 +132,29 @@ export function handleCreateMultisig(event: CreateMultisigWithAgents): void {
...
+
+    // Use the most recently registered agent instead of all agents
+    // This matches the SQL query logic to prevent double counting
+    const mostRecentAgent = getMostRecentAgent(
+      event.params.serviceId.toI32(),
+      service.agentIds,
+      event.block.timestamp
+    );
+
+    if (mostRecentAgent != -1) {
```
Tanya-atatakai, line 148: it's unclear from the function name (getMostRecentAgent) that in case of not finding the agent it'll return -1. can it return null?

shared/service-registry/mappers/mapping.ts
```diff
@@ -125,11 +132,29 @@ export function handleCreateMultisig(event: CreateMultisigWithAgents): void {
...
+      multisig.agentIds = [mostRecentAgent];
```
Tanya-atatakai, line 149: 1) if mostRecentAgent returns agent ID, can we rename the fun to `mostRecentAgentId` please?
2) if there's always one agent per multisig, should we still keep it as an array?

shared/service-registry/utils.ts
```diff
@@ -205,4 +206,55 @@ export function createDailyActiveMultisig(
...
+  const id = serviceId.toString().concat("-").concat(agentId.toString());
+  let registration = AgentRegistration.load(id);
+  if (registration == null) {
+    registration = new AgentRegistration(id);
+    registration.serviceId = serviceId;
+    registration.agentId = agentId;
+    registration.registrationTimestamp = timestamp;
+  }
+  registration.save();
+}
```
Tanya-atatakai, line 225: we only update timestamps when we create the entity, is that correct? shouldn't it be something like
```suggestion
  const id = serviceId.toString().concat("-").concat(agentId.toString());
  let registration = AgentRegistration.load(id);
  if (registration == null) {
    registration = new AgentRegistration(id);
  }
  registration.serviceId = serviceId;
  registration.agentId = agentId;
  registration.registrationTimestamp = timestamp;
  registration.save();
}
```

subgraphs/service-registry/service-registry-eth/src/utils.ts
```diff
@@ -205,4 +202,55 @@ export function createDailyActiveMultisig(
...
+  if (registration == null) {
+    registration = new AgentRegistration(id);
+    registration.serviceId = serviceId;
+    registration.agentId = agentId;
+    registration.registrationTimestamp = timestamp;
+  }
+  registration.save();
+}
+
+export function getMostRecentAgent(
```
Tanya-atatakai, line 223: this function looks completely identical with the one in `shared/service-registry/utils.ts` - can we just reuse it?

shared/service-registry/mappers/mapping.ts
axatbhardwaj: 1. renamed 
2. it is used for historical accuracy

subgraphs/service-registry/service-registry-eth/src/utils.ts
Tanya-atatakai: from what I see, everything can be shared for both l1 and l2, including schema. the only difference is in CreateService handler, which can be separated, but still inside it we can reuse the same getOrCreateService

making this copy pasted will cause issues in future. if there's no time for it, please add a ticket to backlog.

PR #45, opened 2025-10-02
Feature/ve olas locked

axatbhardwaj:
### Summary

- Implement DepositorLock + WeeklyDepositorsUnlock (derived) to track “currently locked” depositors without FE filtering.
- Add Global.activeLockedHolderCount; update counters on true state transitions.
- Block handler expires locks by scanning current and previous week; no arrays stored, derived lists used on demand.

## Key changes

Add/update:

- DepositorLock: account, unlockTimestamp, isActive, isLocked, weeklyUnlock
- WeeklyDepositorsUnlock: id=weeklyStartTimestamp (derived locks)
- Global: veolasHolderCount, activeLockedHolderCount, updatedAt


subgraphs/tokenomics/tokenomics-eth/src/veolas.ts
```diff
@@ -1,49 +1,127 @@
...
   if (depositor != null && depositor.isActive) {
     depositor.isActive = false;
     depositor.save();
+  }
+
+  let depositorLock = loadOrCreateDepositorLock(event.params.account);
+
+  if (!depositorLock.isActive) {
+    return;
+  }
```
Tanya-atatakai, line 62: I don't think this is possible, smart contract should prevent from this

subgraphs/tokenomics/tokenomics-eth/src/veolas.ts
```diff
@@ -1,49 +1,127 @@
...
+
+  let globalMetrics = getOrCreateGlobalMetrics();
+  const becameLocked = depositorLock.isLocked && !wasLocked;
+  incrementGlobalCountersForDeposit(
+    globalMetrics,
+    wasInactive,
+    becameLocked,
+    event.block.timestamp
+  );
+  globalMetrics.save();
```
Tanya-atatakai, line 48: why can't we do getOrCreateGlobalMetrics and entity save right inside incrementGlobalCountersForDeposit? the fn name sounds like it should do it

subgraphs/tokenomics/tokenomics-eth/src/veolas.ts
```diff
@@ -1,49 +1,127 @@
...
+
+  depositorLock.save();
+
+  let globalMetrics = getOrCreateGlobalMetrics();
+  decrementGlobalCountersForWithdraw(
+    globalMetrics,
+    wasLocked,
+    event.block.timestamp
+  );
+  globalMetrics.save();
```
Tanya-atatakai, line 76: came proposal do it it all right inside decrementGlobalCountersForWithdraw
wdyt?

subgraphs/tokenomics/tokenomics-eth/src/veolas.ts
```diff
@@ -1,49 +1,127 @@
...
+}
+
+export function handleBlock(block: ethereum.Block): void {
+  const currentTimestamp = block.timestamp;
+
+  const currentWeekStart = getCurrentWeekStart(currentTimestamp);
+  const WEEK_SECONDS = BigInt.fromI32(7 * 24 * 60 * 60);
```
Tanya-atatakai, line 83: can you export the same value from utils?

subgraphs/tokenomics/tokenomics-eth/src/veolas.ts
```diff
@@ -1,49 +1,127 @@
...
+  const weekStarts: BigInt[] = [currentWeekStart, currentWeekStart.minus(WEEK_SECONDS)];
```
Tanya-atatakai, line 84: why should we handle past week in current block?

subgraphs/tokenomics/tokenomics-eth/src/veolas.ts
```diff
@@ -1,49 +1,127 @@
...
+    }
+    const expiredLocks = getExpiredLocks(locks, currentTimestamp);
+    if (expiredLocks.length == 0) {
+      continue;
+    }
+    processExpiredLocksBatch(expiredLocks, currentTimestamp);
+  }
+}
+
+function processExpiredLocksBatch(
```
Tanya-atatakai, line 99: could we move it to utils maybe?

subgraphs/tokenomics/tokenomics-eth/schema.graphql
```diff
@@ -471,5 +471,21 @@ type VeolasDepositor @entity(immutable: false) {
...
+  account: Bytes! 
+  unlockTimestamp: BigInt! 
+  isActive: Boolean! 
+  isLocked: Boolean! 
+  weeklyUnlock: WeeklyDepositorsUnlock 
+}
+
+type WeeklyDepositorsUnlock @entity(immutable: false) {
+  id: ID!
+  weeklyStartTimestamp: BigInt!
```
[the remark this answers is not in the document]
axatbhardwaj: startTimestamp it is ,done

subgraphs/tokenomics/tokenomics-eth/src/veolas.ts
axatbhardwaj: - To catch boundary expiries. If a lock expires in the last seconds of a week and there’s no block after that timestamp before the week rolls, the first block that can see it is in the next week. Checking the previous week ensures we don’t miss those.
- It’s cheap. We load the previous week only if it exists; if empty, we skip immediately.

subgraphs/tokenomics/tokenomics-eth/src/veolas.ts
axatbhardwaj: Done

subgraphs/tokenomics/tokenomics-eth/src/veolas.ts
Tanya-atatakai: ok could you please add a small comment re this, otherwise it's a bit confusing like there's a specific business logic for doing this

Tanya-atatakai (approved):

rajat2502 (approved):

PR #50, opened 2025-10-03
Feat/daily staking global totalrewards

axatbhardwaj:
For BabyDegen Subgraph metrics we need , rewards till a date (cumulative rewards earned by agents till that day) - we need it for last 7 days

- Track global.totalRewards per checkpoint and persist a DailyStakingGlobal record at day start.
- Added helpers dayStart and upsertDailyStakingGlobal to keep logic clean.
- Updated codegen-staking to emit bindings under subgraphs/staking/generated.

### DIFF
--- a/subgraphs/staking/src/staking-proxy.ts
+++ b/subgraphs/staking/src/staking-proxy.ts
@@ -9,7 +9,7 @@ import {
   ServiceUnstaked as ServiceUnstakedEvent,
   ServicesEvicted as ServicesEvictedEvent,
   Withdraw as WithdrawEvent,
-} from "../generated/templates/StakingProxy/StakingProxy"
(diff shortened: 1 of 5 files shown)


DIvyaNautiyal07 (approved):

subgraphs/staking/src/staking-proxy.ts
```diff
@@ -55,6 +55,15 @@ export function handleCheckpoint(event: CheckpointEvent): void {
     }
   }
 
+  // Update global cumulative staking rewards
+  let global = getGlobal();
+  global.totalRewards = global.totalRewards.plus(totalRewards);
+  global.save();
```
axatbhardwaj, line 61: @Tanya-atatakai 

this should be the fix for global entity right ?

PR #74, opened 2025-12-29
PRO-849 : Terminated services Handling

axatbhardwaj:
Problem
The Scenario:

1. Service Terminated: agentIds are cleared [], but multisig remains set to the old address (this is the on-chain behavior).
2. Service Re-activated: ActivateRegistration event is emitted on-chain, but the subgraph has no handler for it.
3. Instances Registered: RegisterInstance events are emitted.

The Result in Subgraph:

- agentIds: Contains the new agent IDs from the new registration
- multisig: Still points to the previous (old) multisig address
- The multisig field is only updated when deploy is called again, emitting CreateMultisigWithAgents.

Additional Issue:

If the old multisig makes transactions after terminate but before new deploy, those transactions would still be tracked in DailyServiceActivity and agent performance metrics — associating them with a terminated service.
The Fix
1. Added ServiceState enum to the Service entity to mirror on-chain states:
   - PreRegistration, ActiveRegistration, FinishedRegistration, Deployed, TerminatedBonded
2. Track state transitions in handlers:

    - handleCreateService → PreRegistration
    - handleCreateMultisig → Deployed
    - handleTerminateService → TerminatedBonded

3. Added state check in execution handlers:

    - handleExecutionSuccess and handleExecutionFromModuleSuccess now skip activity tracking if service.state != "Deployed"
    - Prevents old multisig transactions from creating DailyServiceActivity entries for terminated services

Result:
On-chain parity maintained (multisig reference is NOT cleared on terminate)
Consumers can filter by state to distinguish service lifecycle stages
Old multisig activity after termination is not tracked against the service

### DIFF
--- a/subgraphs/service-registry/schema.graphql
+++ b/subgraphs/service-registry/schema.graphql
@@ -1,9 +1,18 @@
+enum ServiceState {
+  PreRegistration
+  ActiveRegistration
+  FinishedRegistration
+  Deployed
+  TerminatedBonded
+}
+
 type Service @entity(immutable: false) {
   id: ID! # service ID
   multisig: Bytes
   agentIds: [Int!]!
   creationTimestamp: BigInt!
   configHash: Bytes
+  state: ServiceState!
 }
 
 type AgentRegistration @entity(immutable: false) {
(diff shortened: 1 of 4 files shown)


axatbhardwaj:
Closing this for now — you can pick things up later when ready. cc @mariapiamo @rajat2502 @Tanya-atatakai

PR #121, opened 2026-05-07
chore: supply chain update - 3

mohandast52:
## Summary

Phase 3 (final) of the supply-chain hardening rollout (see [`.plans/supply-chain-update.md`](.plans/supply-chain-update.md), Tier 2 + parts of Tier 4). **Stacked on #120, which stacks on #119.** Will rebase to `main` once those merge.

Adds the regression-prevention layer: every future PR is automatically audited for new vulnerabilities, new install hooks, lockfile drift, and leaked secrets. **All new gates land advisory-only** — branch protection isn't being changed here, so failing gates won't block merging until you promote them in repo Settings.

Closes T5 (future-leak detection) from the threat model.

### Workflows (added)
- [`.github/workflows/supply-chain.yml`](.github/workflows/supply-chain.yml) — runs `yarn audit:prod` + install-hook gate at root, plus `lockfile-lint` matrixed over all 13 paths, with `all-checks-passed` aggregator job.
- [`.github/workflows/gitleaks.yml`](.github/workflows/gitleaks.yml) — secret scan on every push + PR, with **SHA-256 verified gitleaks 8.30.1 binary** (`551f6fc83ea457d62a0d98237cbad105af8d557003051f41f3e7ca7b3f2470eb`, cross-checked against townhall-kpis and upstream `checksums.txt`).

### Scripts (added)
- [`scripts/audit.mjs`](scripts/audit.mjs) — wraps `yarn audit --json` with allowlist suppression. **Critical: exposed as `yarn audit:prod`, NOT `yarn audit`** (yarn 1.x's built-in shadows same-named scripts in package.json). Time-bounded suppression: each allowlist entry needs `id`, `reason`, `added`, `review` (all required); expired entries print a CI warning but don't fail.
- [`scripts/audit-install-hooks.mjs`](scripts/audit-install-hooks.mjs) — enumerates packages in `node_modules` declaring non-trivial pre/install/postinstall scripts, diffs against [`.supply-chain/install-hooks.allowlist`](.supply-chain/install-hooks.allowlist). Drift in **either direction** (new hook OR removed hook) fails the job — the latter catches stale allowlist entries.

### Allowlists (added, baselined)
- [`.supply-chain/audit-allowlist.json`](.supply-chain/audit-allowlist.json) — **23 unique high/critical advisories** baselined as transitive of `@graphprotocol/graph-cli` (latest stable `0.98.1` doesn't refresh them). 90-day review cadence.
[...]

PR #122, opened 2026-05-07
chore: supply chain update - 4

mohandast52:
## Summary

Follow-up to #121: clear 3 of the 23 audit allowlist entries via yarn `resolutions`. Conservative scope — only resolutions where I'm 100% confident no consumer breaks. **Stacked on #121.**

| Before | After |
|---|---|
| 23 allowlisted advisories | **20 allowlisted advisories** (3 cleared) |

## Changes

Single resolutions block added to root `package.json`:

```json
"resolutions": {
  "immutable": "^5.1.5",
  "cross-spawn": "^7.0.6",
  "semver": "^7.5.2"
}
```

Root `yarn.lock` regenerated. 3 entries removed from `.supply-chain/audit-allowlist.json`.

Net diff: 3 files, +12 / −71 lines.

## Why these three (and not others)

| Package | Versions in tree | Decision | Reason |
|---|---|---|---|
| `immutable` | 5.1.4 only | **Resolved → ^5.1.5** | Single major in tree, in-major patch bump |
| `cross-spawn` | 7.0.3 + 7.0.6 | **Resolved → ^7.0.6** | Single major; just dedupes the two patches |
| `semver` | 7.3.5 + 7.7.2/3/4 | **Resolved → ^7.5.2** | Single major; 7.3.5 forwards to 7.5.2+ |
| `picomatch` | 2.3.1 **+ 4.0.4** | Skipped | Multi-major — forcing one would break the other's consumers |
| `minimatch` | 3.1.2 **+ 5.1.6 + 9.0.5 + 10.2.5** | Skipped | 4 majors in tree — high risk of consumer break |
| `glob` | 7.2.3 **+ 11.0.3** | Skipped | API differs significantly between 7.x and 11.x |
| `axios` (5 advisories!) | 0.21.4 | Skipped | Needs migration to ≥0.30 / 1.x; graph-cli uses 0.21 internally |
| `lodash` | 4.17.21 | Skipped | Advisory affects ALL of 4.x; no fix without major migration to 5.x |
| `undici` (3 advisories) | 7.16.0 | Skipped | Needs investigation against the actual fix target |

The skipped ones are tracked for the August 2026 quarterly review or for the eventual graph-cli upstream bumps.

### Root (where the resolution is applied)

| Check | Result |
|---|---|
| `yarn install --frozen-lockfile` after edit | clean — Saved lockfile, deterministic |
| Resolution actually took effect | `immutable: 5.1.5`, `cross-spawn: 7.0.6` (deduped), `semver: 7.7.4` (forced 7.3.5 forward) |
| `yarn audit:prod` | OK — 20 allowlisted, no unlisted high/critical |
| Drift detector | Correctly flagged 3 stale entries before prune; quiet after prune |

### All 13 paths still deterministic

```
.                          success Already up-to-date.
governance                 success Already up-to-date.
legacy-mech-fees           success Already up-to-date.
liquidity                  success Already up-to-date.
liquidity-l2               success Already up-to-date.
new-mech-fees              success Already up-to-date.
service-registry           success Already up-to-date.
staking                    success Already up-to-date.
tokenomics-eth             success Already up-to-date.
tokenomics-l2              success Already up-to-date.
predict-omen               success Already up-to-date.
predict-polymarket         success Already up-to-date.
babydegen-optimism         success Already up-to-date.
```

Subgraph lockfiles are unaffected — root resolutions don't propagate, by design.

### Subgraph smoke tests (representative coverage)

| Subgraph | Result |
|---|---|
| predict-omen | **19/19 tests passed** (same as PR2 verification) |
| babydegen-optimism | **10/10 tests passed** (same as PR2 verification) |

## Why not also fix the multi-major cases

Yarn 1.x `resolutions` are package-name-level — they pin ALL instances of a package to one version. With multi-major packages like picomatch (2.x callers + 4.x callers) or minimatch (4 majors), forcing one version would either:
- Make yarn refuse the install (semver conflict), or
- Force consumers expecting 4.x to use 2.x (or vice versa) — likely runtime break.

Yarn 1.x's selective resolution syntax (`some-parent/picomatch`) requires knowing every parent in advance and tracking new ones over time — high maintenance cost for marginal advisory clearance. Better to wait for upstream to refresh those transitives.

## Test plan

- [x] Resolutions applied; `yarn install --frozen-lockfile` clean.
- [x] Audit gate now green at 20 allowlisted (was 23).
- [x] All 13 path lockfiles deterministic (subgraph lockfiles untouched).
- [x] predict-omen and babydegen-optimism: 29/29 tests pass.
- [ ] **Reviewer:** push triggers all 4 workflows (test, supply-chain, gitleaks, deploy dispatch). All should pass.

## Rollback

Revert is a single commit. Removes the resolutions block, regenerates yarn.lock, restores 3 allowlist entries. Zero side effects in subgraphs (their lockfiles never changed).

## Stacked on

Base = `mohan/supply-chain-update-3`. Once #121 merges, this PR auto-rebases to `main`.


atepem:
# PR Review: `mohan/supply-chain-update-3` → `mohan/supply-chain-update-4`

Single commit `fc2ae9a` — clears 3 advisories via Yarn `resolutions`, removes the matching allowlist entries.

## What changed

| File | Change |
|---|---|
| `package.json` | `+resolutions` block: `immutable ^5.1.5`, `cross-spawn ^7.0.6`, `semver ^7.5.2` |
| `.supply-chain/audit-allowlist.json` | -3 entries (1117068 immutable, 1112921 semver, 1104664 cross-spawn) |
| `yarn.lock` | Consolidates 4 `semver@*` entries → `7.7.4`; bumps `immutable` 5.1.4→5.1.5; collapses `cross-spawn@7.0.3`→7.0.6; drops dead `lru-cache@6` + `yallist@4` (orphaned by the semver bump) |

## Strengths

- **1:1 pairing** — every resolution is matched by an allowlist removal. Net reduction in suppressions, no orphan entries left behind. The "stale entry" warning in `scripts/audit.mjs:153-157` confirms the script would catch any forgotten removal.
- **Resolutions block carries the rationale** — the inline `//` comment explains *why* picomatch/minimatch/glob were skipped (multiple majors in tree). Future-you (or another reviewer) won't have to re-derive that.
- **Caret floors pinned at the fix version, not above** — `^7.5.2` for semver, even though the installed version is 7.7.4. Correct: pinning the floor at the actual fix lets the lockfile pick a fresher version freely without making the resolution itself the gate.
- **No install-hook drift expected** — `immutable`, `cross-spawn`, `semver`, `lru-cache`, `yallist` are all script-free; `.supply-chain/install-hooks.allowlist` is empty and should stay empty.
- **Lockfile is genuinely smaller** — 71 lines removed vs. 12 added. The consolidation is real, not just rewriting.

## Concerns

### 1. Scope: subgraph yarn.locks still ship the vulnerable versions (worth addressing or calling out)

The supply-chain audit job (`.github/workflows/supply-chain.yml:16-34`) runs `yarn audit:prod` **once, at root** — its name says so explicitly: `"Dependency audit (root tree)"`. The 12 subgraph `yarn.lock` files still contain the unpatched versions:

```
subgraphs/*/yarn.lock:  immutable@5.1.4 / cross-spawn@7.0.3 / semver@7.3.5
```

These are the lockfiles consumed when each subgraph runs `yarn install` for codegen/build/test/deploy. The PR description ("clear 3 audit advisories") is technically accurate for the gated tree, but the same vulnerable code still lands in every subgraph's `node_modules`. This is a pre-existing gap (introduced when supply-chain CI was scoped to root in update-3), not something this PR caused — but since the resolutions exist now, it'd be cheap to propagate them. Two options:

- **Fix here**: copy `resolutions` into all 12 subgraph package.jsons + `yarn install` in each to refresh locks. Mechanical, but 24 file changes.
- **Defer + document**: leave a note in `SUPPLY-CHAIN-SECURITY.md` that the audit gate covers root only and subgraph trees aren't currently audited.

I'd lean toward the first — same advisories, same fix, the work is pure copy-paste, and it eliminates the divergence between what CI gates and what actually deploys.

### 2. `"//": "..."` key inside `resolutions` may trip Yarn 1

Yarn 1 treats every key in `resolutions` as a package selector and emits a warning (sometimes an error in `--frozen-lockfile` mode) for unknown selectors. The standard JSON-comment hack works in `dependencies` because Yarn ignores unresolved deps, but `resolutions` is parsed differently. Worth a quick local check:

```bash
yarn install --frozen-lockfile
```

If it warns or fails, move the comment out of `resolutions` — either inline above the block as a regular `"//"` field at the top level of `package.json`, or into `SUPPLY-CHAIN-SECURITY.md` where the rest of the supply-chain rationale already lives. (The doc-in-the-file rationale is genuinely useful — I'd rather see it relocated than dropped.)

### 3. Minor: the comment lists the multi-major exclusions but the allowlist still references them

The remaining allowlist entries for picomatch/minimatch/glob/etc. don't cite the "multiple majors" reasoning — they all say `"latest stable graph-cli (0.98.1) does not refresh this advisory"`. Both reasons are true, but a future reviewer trying to clear those entries with another resolution will only see the upstream-bump reason and miss the more fundamental "can't, would break installs" constraint that's documented over in package.json. Optional: cross-reference the package.json comment from those allowlist entries' `reason` field, or add the multi-major note to those entries.

## Verdict

The mechanism is sound and the diff is clean. **#2 is the one I'd want resolved before merge** (5-second check), **#1 is a judgment call** about scope creep vs. closing a gate-vs-reality gap, **#3 is polish**.

Tanya-atatakai (approved):
LGTM. Clean, conservative supply-chain cleanup.

Verified:
- Allowlist now at 20 entries (was 23); the 3 removed GHSAs map 1:1 to the 3 added resolutions.
- Resolutions take effect in yarn.lock: immutable→5.1.5, cross-spawn deduped to 7.0.6, semver consolidated to 7.7.4. Stale transitives (lru-cache@^6, yallist@^4) correctly pruned.
- Caret floors are fix-version minima, not arbitrary picks.
- Multi-major guardrail captured inline via the `//` comment in package.json — exactly where a future contributor would be tempted to add picomatch/minimatch/glob.
- Rollback is a single revert; subgraph lockfiles untouched by design.

Risk: low. Dep-only, scope is well-defended, CI drift detector already proved itself by flagging the stale entries.

mohandast52:
Thanks @atepem — solid review, addressed all three points. New commit: [`e7f8b62`](https://github.com/valory-xyz/autonolas-subgraph-studio/commit/e7f8b62).

## Concern #1 — subgraph yarn.locks shipped vulnerable versions ✅ Fixed

Confirmed your point: all 12 subgraph yarn.locks had `immutable@5.1.4`, `cross-spawn@7.0.3`, `semver@7.3.5`. The root audit gate said one thing, the deploy surface said another.

Picked **option A** (propagate). Mirrored the root `resolutions` block into all 12 subgraph `package.json`s (with a short pointer back to root for the rationale), then ran `yarn install` per subgraph to refresh each `yarn.lock`. After-state across **all 12** subgraphs:

| Package | Before | After |
|---|---|---|
| `immutable` | 5.1.4 | **5.1.5** |
| `cross-spawn` | 7.0.3 | **7.0.6** |
| `semver` | 7.3.5 | **7.7.4** (forced via `^7.5.2` floor) |

Verified with full local sequence on three CI patterns:

| Subgraph | Pattern | Result |
|---|---|---|
| `legacy-mech-fees` | single-network | codegen + build + **15/15** tests |
| `predict-omen` | nested | codegen + build + **19/19** tests |
| `predict-polymarket` | nested, dep-heavy | codegen + build + **96/96** tests |
| | total | **130/130** ✅ |

Plus `yarn install --frozen-lockfile` clean on **all 12** (12/12).

## Concern #2 — `"//"` key inside `resolutions` may trip Yarn 1 ✅ No change needed (verified empirically)

Tested locally on Node 22.18.0 + Yarn 1.22.22:

```
$ yarn install --frozen-lockfile
yarn install v1.22.22
[1/5] Validating package.json...
[2/5] Resolving packages...
[3/5] Fetching packages...
[4/5] Linking dependencies...
warning "@graphprotocol/graph-cli > web3-eth-abi > abitype@0.7.1" has unmet peer dependency "typescript@>=4.9.4".
[5/5] Building fresh packages...
Done in 0.83s.
```

Only warning is the unrelated `typescript` peer dep. **No complaint about the `//` key in `resolutions`** — Yarn 1 silently accepts unknown selectors there. Doc-in-the-file rationale stays put.

## Concern #3 — allowlist entries cite only the graph-cli ceiling ✅ Fixed

Appended the multi-major reasoning to all 11 picomatch/minimatch/glob entries. Now a future reviewer attempting to clear those entries with another resolution sees both reasons:

> Baselined at PR3 — transitive of @graphprotocol/graph-cli; latest stable graph-cli (0.98.1) does not refresh this advisory. Re-evaluate when graph-cli is bumped or a yarn resolution becomes feasible. **A scoped Yarn `resolutions` would force one major across all consumers and break installs (multi-major in tree).**

## Drive-by

The npm advisory DB rotated `1113274` → `1117857` for the same axios CVE (`GHSA-43fc-jf86-j433`). Root `yarn audit:prod` was failing with *"stale entry"* on `1113274` + *"unlisted"* on `1117857`. Replaced the id in place; reason / dates / GHSA preserved. Final root state: **20 allowlisted, 0 unlisted, exit 0.**

## Diff summary

`25 files changed, 227 insertions, 486 deletions` — most of the deletions come from yarn.lock dedup (semver consolidation across paths, dropped orphans like `lru-cache@6` and `yallist@4`), same shape you flagged in the original PR description.

rajat2502 (approved):
LGTM. Mechanism is sound, scope is well-defended, and @atepem's three concerns (subgraph propagation, `"//"` key safety, multi-major rationale on remaining entries) are all addressed in `e7f8b62`.

Verified:
- 1:1 pairing between resolutions added and allowlist entries removed.
- Caret floors pinned at the documented fix version (`^7.5.2` / `^7.0.6` / `^5.1.5`), letting the lockfile pick fresher minors freely.
- Subgraph mirrors close the gate-vs-reality gap — the audit workflow only covers root, but the deploy surface (12 subgraph trees) now ships the patched versions too.
- Smoke tests across single-network / nested / dep-heavy patterns (legacy-mech-fees, predict-omen, predict-polymarket = 130/130) cover the riskiest bump (semver 7.3.5 → 7.7.4).
- axios advisory id rotation (1113274 → 1117857) preserves the GHSA — same CVE under a renumbered npm record.

Non-blocking follow-ups for whoever picks up #5:
- Add a CI check that diffs root vs. subgraph `resolutions` so the next root-level addition can't silently skip the 12 mirrors.
- Consider factoring the repeated "multi-major-blocked" rationale into the allowlist schema instead of free-text duplication across 11 entries.

PR #124, opened 2026-05-13
chore: dependabot update - 1

mohandast52:
## Summary

Add Yarn `resolutions` for 6 transitive dependencies to clear **~132 Dependabot alerts** (49 high, 83 medium/low) across the 13 `yarn.lock` files in the monorepo. All targets are **single major in tree** and live in CLI tooling paths only (`@graphprotocol/graph-cli` internals) — no subgraph runtime code is touched.

### Resolutions (root + 12 subgraph `package.json`, mirrored)

| Package | Pre-bump (in tree) | Resolution range | Resolved version | Why |
|---|---|---|---|---|
| `lodash` | 4.17.21 | `^4.18.0` | 4.18.1 | Clears 3 advisories (10 high) — utility used by `@oclif/plugin-warn-if-update-available` |
| `js-yaml` | 4.1.0 | `^4.1.1` | 4.1.1 | Clears 1 advisory (13 med) — patch bump (4.1.0 → 4.1.1) |
| `follow-redirects` | 1.14.0 | `^1.16.0` | 1.16.0 | Clears 2 advisories (11 med) — minor bump |
| `ejs` | 3.1.8 + 3.1.10 | `^3.1.10` | 3.1.10 | Clears 1 advisory (13 med) — also converges duplicate range |
| `tmp` | 0.2.0 | `^0.2.4` | 0.2.5 | Clears 1 advisory (3 low) — patch bump |
| `undici` | 7.16.0 | `^7.24.0` | 7.25.0 | Clears 6 advisories (39 high, 39 med) — minor bump within 7.x |

### Allowlist drift cleanup

[`.supply-chain/audit-allowlist.json`](.supply-chain/audit-allowlist.json): **20 → 16 entries**. Removed entries that no longer suppress any high/critical advisory (confirmed by `yarn audit:prod` stale warnings):

- `1115806` (lodash, GHSA-r5fr-rjxr-66jc)
- `1114639` (undici, GHSA-v9p9-hfj2-hcw8)
- `1114637` (undici, GHSA-vrm6-8vpv-qv8q)
- `1114591` (undici, GHSA-f269-vfmq-vjvj)

### Docs

- [`SUPPLY-CHAIN-SECURITY.md`](SUPPLY-CHAIN-SECURITY.md) §11: document that `graph-cli` HTTP transitive bumps (`undici`, `axios`) require a `workflow_dispatch` staging deploy gate before production. Codifies the recommendation below as a repeatable process gate.

## Why these and not others?

Per the rationale already in [`package.json`](package.json)'s `resolutions` block, a Yarn resolution is only safe when the target package has **one major version** in the dep tree — otherwise it forces one major across consumers expecting another and breaks installs.

**Eligible for resolution (this PR):** lodash, js-yaml, follow-redirects, ejs, tmp, undici — all single-major.

**Still allowlisted (16 entries), cannot be resolved:**
- `minimatch` (3.x + 5.x + 9.x + 10.x in tree)
- `picomatch` (2.x + 4.x)
- `glob` (7.x + 11.x)
- `brace-expansion` (1.x + 2.x + 5.x)
- `axios` — single major, but sits on `graph-cli`'s deploy HTTP path via `apisauce`. Deserves a separate PR with a staging-deploy gate. **~195 axios alerts (65 high) will be the target of Tier 2B.**

## Test plan

All run locally on Node 24.15.0 / Yarn 1.22.22 before push:

- [x] `yarn install --frozen-lockfile` on root + 12 subgraphs
- [x] `yarn graph codegen` on all 12 subgraphs
- [x] `yarn graph test` (Matchstick): **258 / 258 tests pass**
- [x] `yarn graph build` validated on 3 subgraph shapes (standalone: liquidity, template: staking, per-network: new-mech-fees) — produces deploy artifact successfully. CI's matrix only runs codegen+test, so build was tested locally for the deploy path.
- [x] `yarn audit:prod`: clean (16 allowlisted, 0 unlisted high/critical, no expired entries)
- [x] `yarn audit:install-hooks`: clean (no install-hook drift)
- [x] `lockfile-lint` on all 13 `yarn.lock` files: pass

CI re-runs codegen + test for all 12 subgraphs, plus audit + install-hooks + lockfile-lint + gitleaks via the [consolidated `ci.yml`](.github/workflows/ci.yml). All checks green.

## Risk assessment

- **Subgraph runtime / indexed data**: zero risk. None of the 6 bumped packages are reachable from compiled WASM. `graph build` validates artifact integrity.
- **`graph deploy` path**: ~98% confidence. `undici` is used by `graph-cli` for HTTP upload to The Graph Studio. The 7.16 → 7.25 bump is same-major, security-only fixes per the upstream changelog, but can't be fully verified without an actual deploy.
- **Rollback**: `git revert <merge-sha>`. Symmetric — bumping resolutions back is one-line per package.

## Pre-production-deploy gate (required, now documented)

Per [`SUPPLY-CHAIN-SECURITY.md`](SUPPLY-CHAIN-SECURITY.md) §11: after merge, trigger a `workflow_dispatch` deploy of `legacy-mech-fees` (smallest, single-chain) to a staging slug from the merge commit before any production deploy. This validates the `undici` upload path end-to-end.

## Out of scope

- **Tier 2B — axios resolution** (~195 alerts, 65 high). Separate PR with the same staging-deploy gate.
- **Tier 3 — multi-major packages** (`minimatch`, `picomatch`, `glob`, `brace-expansion`). Resolved only when `@graphprotocol/graph-cli` bumps its own transitives. Tracked via the 16 remaining `audit-allowlist.json` entries with `review: 2026-08-05`.

### DIFF
--- a/.supply-chain/audit-allowlist.json
+++ b/.supply-chain/audit-allowlist.json
@@ -1,13 +1,13 @@
 {
-  "_doc": "Allowlist for yarn audit high/critical advisories. Every entry needs a reason and a review date. See SUPPLY-CHAIN-SECURITY.md §5.",
+  "_doc": "Allowlist for yarn audit high/critical advisories. Every entry needs a reason and a review date. See SUPPLY-CHAIN-SECURITY.md \u00a75.",
   "_fields": {
-    "id": "npm advisory numeric ID — REQUIRED",
-    "ghsa": "GitHub Security Advisory ID — for humans",
-    "package": "package name — for humans",
-    "severity": "advisory severity — for humans",
-    "reason": "why the advisory is allowlisted — REQUIRED",
-    "added": "YYYY-MM-DD — REQUIRED",
-    "review": "YYYY-MM-DD — REQUIRED, expired prints warning but does not fail"
+    "id": "npm advisory numeric ID \u2014 REQUIRED",
+    "ghsa": "GitHub Security Advisory ID \u2014 for humans",
+    "package": "package name \u2014 for humans",
+    "severity": "advisory severity \u2014 for humans",
+    "reason": "why the advisory is allowlisted \u2014 REQUIRED",
+    "added": "YYYY-MM-DD \u2014 REQUIRED",
+    "review": "YYYY-MM-DD \u2014 REQUIRED, expired prints warning but does not fail"
   },
   "entries": [
     {
@@ -15,7 +15,7 @@
       "ghsa": "GHSA-6chq-wfr3-2hj9",
       "package": "axios",
       "severity": "high",
-      "reason": "Baselined at PR3 — transitive of @graphprotocol/graph-cli; latest stable graph-cli (0.98.1) does not refresh this advisory. Re-evaluate when graph-cli is bumped or a yarn resolution becomes feasible.",
+      "reason": "Baselined at PR3 \u2014 transitive of @graphprotocol/graph-cli; latest stable graph-cli (0.98.1) does not refresh this advisory. Re-evaluate when graph-cli is bumped or a yarn resolution becomes feasible.",
       "added": "2026-05-07",
       "review": "2026-08-05"
     },
@@ -24,7 +24,7 @@
       "ghsa": "GHSA-pf86-5x62-jrwf",
       "package": "axios",
       "severity": "high",
-      "reason": "Baselined at PR3 — transitive of @graphprotocol/graph-cli; latest stable graph-cli (0.98.1) does not refresh this advisory. Re-evaluate when graph-cli is bumped or a yarn resolution becomes feasible.",
+      "reason": "Baselined at PR3 \u2014 transitive of @graphprotocol/graph-cli; latest stable graph-cli (0.98.1) does not refresh this advisory. Re-evaluate when graph-cli is bumped or a yarn resolution becomes feasible.",
       "added": "2026-05-07",
       "review": "2026-08-05"
     },
@@ -33,16 +33,7 @@
       "ghsa": "GHSA-pmwg-cvhr-8vh7",
       "package": "axios",
       "severity": "high",
-      "reason": "Baselined at PR3 — transitive of @graphprotocol/graph-cli; latest stable graph-cli (0.98.1) does not refresh this advisory. Re-evaluate when graph-cli is bumped or a yarn resolution becomes feasible.",
-      "added": "2026-05-07",
-      "review": "2026-08-05"
-    },
-    {
-      "id": 1115806,
-      "ghsa": "GHSA-r5fr-rjxr-66jc",
-      "package": "lodash",
-      "severity": "high",
-      "reason": "Baselined at PR3 — transitive of @graphprotocol/graph-cli; latest stable graph-cli (0.98.1) does not refresh this advisory. Re-evaluate when graph-cli is bumped or a yarn resolution becomes feasible.",
+      "reason": "Baselined at PR3 \u2014 transitive of @graphprotocol/graph-cli; latest stable graph-cli (0.98.1) does not refresh this advisory. Re-evaluate when graph-cli is bumped or a yarn resolution becomes feasible.",
       "added": "2026-05-07",
       "review": "2026-08-05"
     },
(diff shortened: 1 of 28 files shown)


Tanya-atatakai:
## Review

**Verdict**: Approve with one small cleanup. The mechanical changes are correct, well-scoped, and exhaustively verified locally. One cosmetic regression in the allowlist file is worth fixing before merge.

### Strengths

- **Resolutions perfectly mirrored** across root + 12 subgraph `package.json` files (verified via `jq` — identical key sets and ranges everywhere).
- **Allowlist drift cleanup is correct**: the 4 removed advisory IDs (1115806 lodash, 1114639/1114637/1114591 undici) match the resolved packages.
- **Convention-consistent**: caret ranges in `resolutions` align with the pre-existing `immutable`/`cross-spawn`/`semver` entries, and the rationale comment in root `package.json` correctly documents the one-major-only constraint.
- **Risk reasoning is sound**: targets live in `@graphprotocol/graph-cli` CLI internals only, not in the WASM-reachable mapping code. `graph build` on three representative shapes (standalone, template, per-network) is a reasonable proxy for the deploy artifact.
- **Good staging**: defers `axios` (~195 alerts, on the `graph deploy` HTTP path via `apisauce`) and multi-major packages to dedicated work rather than bundling.

### Issues

**1. Cosmetic regression in `.supply-chain/audit-allowlist.json` — worth fixing before merge**

Every em-dash (`—`) and section sign (`§`) in the file got rewritten to `—` / `§` escape sequences — not just in the entries that changed, but across `_doc`, `_fields`, and every surviving `reason` string. The JSON parses identically so there's no behavioral impact, but:

- The file is meaningfully harder to read on review.
- This is the signature of `json.dumps(..., ensure_ascii=True)` (Python's default) — i.e., a script re-serialized the whole file when it only meant to drop 4 entries. If such a script exists in this workflow, it should use `ensure_ascii=False` (or `jq`) so the file round-trips cleanly.
- Future hand-edits to reason strings will use literal Unicode and look inconsistent with the now-escaped baseline.

Suggested fix: revert the unicode-escaped lines that did not need to change. The structural diff (4 entries removed) is what we actually want to land.

**2. No CI gate for the `graph deploy` HTTP path**

The PR is ~98% confident on the `undici` upgrade because CI runs codegen + test + audit + install-hooks + lockfile-lint, not `graph build` or any deploy dry-run. The "trigger a staging deploy of `legacy-mech-fees` post-merge" follow-up is the right idea but lives in the PR description as a recommendation, not a process gate. Consider documenting this as a required pre-prod-deploy step in `SUPPLY-CHAIN-SECURITY.md` so it doesn't drift away.

**3. Minor PR doc nit**

The "Range / Resolved" table reads `js-yaml: ^4.1.1 / 4.1.1` which looks like a no-op pin, but the lockfile diff shows the resolution actually does work (4.1.0 → 4.1.1). Not blocking — just confusing on a quick read.

### Security

- No new install hooks added — `audit:install-hooks` runs clean per the test plan.
- Resolutions are bounded to a single major (verified by the in-file rationale comment); they cannot accidentally fork transitives into a broken state.
- Allowlist still requires `id`/`reason`/`added`/`review` on the 16 remaining entries — schema discipline preserved.

### Recommendation

Approve once the unicode-escape regression in `audit-allowlist.json` is reverted (or proactively fixed in the script that produced it). Everything else looks good to merge.

mohandast52:
Thanks for the thorough review! All three items addressed in 08b6a7e:

**1. Unicode-escape regression in `audit-allowlist.json`** — fixed. Root cause was exactly as you diagnosed: my one-shot Python script used `json.dumps()` with the default `ensure_ascii=True`. Re-serialized with `ensure_ascii=False` so the file round-trips cleanly. Diff now shows only the 4 structural entry removals; em-dashes and `§` preserved.

**2. Deploy-path CI gate** — added a §11 entry in [`SUPPLY-CHAIN-SECURITY.md`](SUPPLY-CHAIN-SECURITY.md) noting that `graph-cli` HTTP transitive bumps (`undici`, `axios`) require a `workflow_dispatch` staging deploy from the merge commit before any production deploy, since CI only exercises `codegen` + `test` + `build`. Same gate as the existing AssemblyScript runtime bump rule. The PR description now references this as a required follow-up, not a "recommendation".

**3. PR doc nit on `js-yaml`** — clarified the resolutions table to show pre-bump (in tree) versions next to the resolution range and resolved version. `js-yaml` is `4.1.0 → 4.1.1`, not a no-op pin.

Tanya-atatakai (approved):

rajat2502 (approved):
LGTM — supply-chain hygiene PR is well-scoped and well-documented.

One thing worth a follow-up note (non-blocking): the root `yarn.lock` had both `tmp@^0.0.33` (via `external-editor@3.1.0` → `inquirer`) and `tmp@^0.2.0` before this change, so the `"tmp": "^0.2.4"` resolution force-bumps `^0.0.33` across a major boundary (0.0.x → 0.2.x). This is technically the cross-major case the PR's own `//` comment in `package.json` warns against. In practice it's safe because `external-editor` only uses `tmp.fileSync({...}) → {name, fd}`, which is stable across these versions — worth a one-line acknowledgement in the description or §11 so future readers don't have to rediscover this.

Also: `follow-redirects` sits on the same `axios`/HTTP path as `undici`, so consider including it in the §11 deploy-gate list alongside `undici` and `axios`.

The §11 staging-deploy gate on `legacy-mech-fees` covers the `undici` exact-pin override (7.16.0 → 7.25.0), which is the highest-risk piece. As long as that gate runs from the merge commit before any prod deploy, ship it.

PR #126, opened 2026-05-14
chore: dependabot update - 3 (path-scoped resolutions clear remaining 11 GHSAs)

mohandast52:
## Summary

Closes the **192 open Dependabot alerts** remaining after PR #125 (axios). Adds two **path-scoped resolutions** in every package.json (root + 12 subgraphs) and regenerates every yarn.lock to cascade-update other transitives to their latest semver-compatible patches.

## Resolutions added

```json
"**/micromatch/picomatch": "2.3.2",
"**/@graphprotocol/graph-cli/glob": "^11.1.0"
```

These force the vulnerable nested instances to their patched versions **without touching other majors in the tree** — rimraf's nested glob@7 stays at 7.2.3, @oclif/core's tinyglobby/picomatch@4 stays at 4.0.4.

The `**/` prefix is critical (Yarn 1 silently ignores bare `parent/pkg` keys for transitively-nested instances). Plus, deleting every yarn.lock before reinstall is required — the lockfile is sticky.

## What the fresh-lockfile cascade additionally fixed

With fresh resolution, Yarn picks latest-semver-compatible for every transitive, which cleared these too without explicit resolutions:

| Pkg | Before | After |
|---|---|---|
| minimatch 3.x | 3.1.2 | **3.1.5** |
| minimatch 5.x | 5.1.6 | **5.1.9** |
| minimatch 9.x | 9.0.5 | 9.0.9 |
| brace-expansion 1.x | 1.1.11 | **1.1.14** |
| brace-expansion 2.x | 2.0.1 | **2.1.0** |
| yaml 1.x | 1.10.2 | **1.10.3** |
| lodash.trim | 4.5.1 | **4.18.0** |

## Cleared GHSAs (11 unique, 192 alerts)

- **picomatch**: GHSA-c2c7-rcm5-vvqj (HIGH), GHSA-3v7f-55p6-f55p (MED)
- **minimatch**: GHSA-23c5-xmqv-rm74 (HIGH), GHSA-3ppc-4f35-3m26 (HIGH), GHSA-7r86-cg39-jmmj (HIGH)
- **brace-expansion**: GHSA-f886-m6hf-6m8v (MED), GHSA-v6h2-p8h4-qcjw (LOW)
- **yaml**: GHSA-48c2-rrv3-qjmp (MED)
- **lodash.trim**: GHSA-29mw-wpgm-hmr9 (MED)
- **glob**: GHSA-5j98-mcp5-4vw2 (HIGH)
[...]

PR #125, opened 2026-05-13
chore: dependabot update - 2

mohandast52:
## Summary

Bump `apisauce 2.1.6 → 3.2.2` via Yarn `resolutions`. This propagates `axios 0.21.4 → 1.16.0` through the `@graphprotocol/graph-cli` transitive tree, clearing **~195 Dependabot alerts (49+ high)** across the 13 `yarn.lock` files in the monorepo. No subgraph runtime code touched.

This is the **Tier 2B follow-on** to #124 — same approach (Yarn resolutions on transitives), same single-major-in-tree constraint, but with one additional gate (staging deploy) because the affected packages live on the `graph deploy` HTTP upload path.

### Resolutions (root + 12 subgraph `package.json`, mirrored)

| Package | Pre-bump (in tree) | Resolution range | Resolved version | Why |
|---|---|---|---|---|
| `apisauce` | 2.1.6 (vuln via axios 0.21.4) | `^3.2.2` | 3.2.2 | apisauce 3.x is the version designed for axios 1.x. Pulled only by `gluegun` → `@graphprotocol/graph-cli`. |

**Indirect effect**: `axios` flips from `0.21.4` to `1.16.0` (latest patched), because `apisauce@3.2.2` declares `axios ^1.11.0`. No need for a separate `axios` resolution — single resolution propagates.

Why bump `apisauce` and not `axios` directly: `apisauce@2.x` hardpins `axios ^0.21.4` in its own `package.json`. Forcing `axios 1.x` directly would mismatch the consumer's declared contract; bumping `apisauce` to a version *built for* axios 1.x is the clean propagation.

### Allowlist drift cleanup

[`.supply-chain/audit-allowlist.json`](.supply-chain/audit-allowlist.json): **16 → 11 entries**. Removed 5 axios entries that no longer suppress any high/critical advisory (confirmed by `yarn audit:prod` stale warnings, then re-run clean):

- `1117592` (axios, GHSA-6chq-wfr3-2hj9)
- `1117590` (axios, GHSA-pf86-5x62-jrwf)
- `1117575` (axios, GHSA-pmwg-cvhr-8vh7)
- `1117857` (axios, GHSA-43fc-jf86-j433)
- `1111034` (axios, GHSA-jr5f-v2jv-69x6)

Unicode preserved (em-dashes, §) per #124 review feedback.

## Why this approach

[...]

PR #127, opened 2026-05-14
feat(skill): add /triage-security — purpose-built Dependabot triage for npm + AssemblyScript

mohandast52:
## Summary

Adds a new Claude skill at [`.claude/skills/triage-security/SKILL.md`](../blob/mohan/triage-security-dependabot-alerts/.claude/skills/triage-security/SKILL.md) that automates Dependabot-alert triage for this repo. The skill walks every open alert and classifies it by whether the vulnerable package is actually reachable from the deployed WASM artifact (`subgraphs/*/src/**/*.ts`) given AssemblyScript's narrow runtime surface — then acts: applicable findings get a tracking issue, not-applicable findings get dismissed with a closed audit-trail issue, allowlisted advisories are skipped.

The skill is **purpose-built for autonolas-subgraph-studio**, not a fork of any other skill. It encodes the deployment-model assumption (only AS code in `subgraphs/*/src/` ships to WASM; everything else is build-tooling) directly into its classification logic.

## Why this skill exists

The repo accumulates Dependabot alerts faster than they can be hand-triaged (387 alerts on the day this skill was first run). Most are against `graph-cli`'s transitive deps (axios, lodash, picomatch, …) and don't apply to the deployed artifact at all — but proving that case-by-case is tedious and error-prone. The skill mechanizes that classification.

## What it does, per alert

| Classification | Action |
| --- | --- |
| GHSA already in `.supply-chain/audit-allowlist.json` | Skip — maintainer's call is authoritative |
| `@types/*` package | Dismiss `not_used` (types-only, no runtime) |
| Not imported from `subgraphs/*/src/**/*.ts` | Dismiss `not_used` + closed audit issue |
| Imported, CVE preconditions absent + high confidence | Dismiss `inaccurate` + closed audit issue |
| Imported, preconditions absent + moderate/low confidence | Open issue + `needs-human-review` (don't dismiss) |
| Imported, preconditions satisfied | Open tracking issue, leave alert open as canonical record |
| Unclassifiable | Skip + stderr line, never auto-dismiss |
| Non-npm ecosystem (Docker / pip / Actions) | Skip (informational) |

Every dismissal generates a closed audit-trail issue labelled `security-audit`, searchable via `gh issue list --state closed --label security-audit`.

## Validation

The skill's methodology was applied (analytically, not by invoking the skill itself) to triage the 387-alert backlog on this repo. The result: **all 387 alerts can be cleared without dismissal** by combining (1) Dependabot's auto-close of stale axios alerts after PR #125 and (2) two path-scoped resolutions in PR #126.

[**PR #126**](https://github.com/valory-xyz/autonolas-subgraph-studio/pull/126) is the worked example — it ships the exact recipe the skill's "Suggested fix" section recommends:
- `"**/micromatch/picomatch": "2.3.2"` (path-scoped)
- `"**/@graphprotocol/graph-cli/glob": "^11.1.0"` (path-scoped)
- Delete every yarn.lock, reinstall — auto-bumps other transitives

PR #126 cleared 11 unique GHSAs (192 alerts) and emptied `.supply-chain/audit-allowlist.json` (5 entries → 0).

## What the skill encoded that wasn't obvious before

- **`**/parent/child` path-scoped resolutions DO work** in Yarn 1 for multi-major trees — IF the `**/` prefix is included and yarn.lock is wiped before reinstall. The prior `.supply-chain/audit-allowlist.json` entries documented "multi-major resolutions break installs" as their blocker; that's true for unscoped resolutions but not for the `**/`-prefixed form. The skill's Design Notes section captures this finding so future maintainers don't re-learn it.

- **`scope=runtime` is misleading in this tree.** yarn.lock declares every transitive as `runtime`, but graph-cli (the largest direct dep) is build-tooling that never ships. Naive classifiers that trust `scope=runtime → PROD` would misclassify all 387 alerts. The skill explicitly overrides: a package with zero imports anywhere in the working tree is **DEV**, regardless of declared scope.

- **`@types/*` short-circuit.** TypeScript types-only packages never have runtime code; a CVE against `@types/X` is dismissable without further analysis.

## Hard rules in the skill (excerpt)

- Never use `tolerable_risk`, `fix_started`, `no_bandwidth` dismissal reasons — risk-accept is the maintainer's call and lives in `.supply-chain/audit-allowlist.json`.
- Conservative default: when uncertain, open an issue; don't dismiss. A false-positive issue is one `gh issue close` away; a false-negative dismissal hides a real vuln.
- Required side-effect of every dismissal: open a closed audit-trail issue **before** the dismissal PATCH, embed its URL in `dismissed_comment`. The 280-char API limit makes the comment alone insufficient as an audit trail.
- `--rerun-dismissed` mode is strictly read-only — surfaces verdict-drift against historical dismissals without mutating anything.

## How to invoke

```
/triage-security                  # all open alerts
/triage-security --limit 10       # cap on alerts processed
/triage-security --rerun-dismissed   # read-only verdict-drift report
```

Requires `gh auth status` to show `security_events` (or admin) scope on the repo; Dependabot dismissals 403 otherwise.

## Test plan

- [ ] Skill body reviewed end-to-end (was reviewed in conversation prior to commit)
- [ ] Skill paths exist in working tree (`subgraphs/*/src/` resolves to 12 dirs)
- [ ] `.supply-chain/audit-allowlist.json` cross-reference works (verified during PR #126 development)
- [ ] When invoked: pre-creates labels (`security`, `dependabot`, `triage-security`, `needs-human-review`, `security-audit`)
- [ ] When invoked: dedupes by GHSA on repeat runs
- [ ] Node-engine fallback documented (§0.4b) — direct `node scripts/audit.mjs` invocation works on node < 24

## Commit hygiene

This branch has three commits with auto-generated messages from the development process. Suggest **squashing on merge** so the final history has one clean commit using this PR's title.

## Recommended merge

"Squash and merge" using this PR's title as the squash-commit message.


.claude/skills/triage-security/SKILL.md
```diff
@@ -0,0 +1,976 @@
...
+fi
+
+# Build a regex that matches both single- and double-quoted imports of the
+# package, including deep subpaths (`axios/dist/foo` still attributes to axios)
+# but anchored at a word boundary so `axios-mock-adapter` doesn't match `axios`.
+PKG_RE=$(printf '%s' "$PKG" | sed 's/[][\\/.*^$+?{}()|]/\\&/g')
+
+# Match: import x from 'pkg' | import 'pkg' | import x from 'pkg/sub' |
+#        require('pkg') | require('pkg/sub') | import('pkg')
+IMPORT_RE="(import[[:space:]]+([^;'\"]*[[:space:]]+from[[:space:]]+)?['\"]${PKG_RE}(/[^'\"]*)?['\"]|require\\([[:space:]]*['\"]${PKG_RE}(/[^'\"]*)?['\"]|import\\([[:space:]]*['\"]${PKG_RE}(/[^'\"]*)?['\"])"
```
DhairyaPatel7: IMPORT_RE is grep-line-anchored: `import\s+...\s+from\s+['"]pkg['"]` only matches when `import` and `from "pkg"` are on the same line. Multi-line imports are used in this repo's AS sources — verified in `subgraphs/babydegen/babydegen-optimism/src/balancerVault.ts`, which has `import {\n  Address,\n  BigInt,\n  ...\n} from "@graphprotocol/graph-ts"`. grep cannot match across lines without `-z`/`-P` multiline mode, so PROD_HIT_FILES omits these files. Per the §2.4 table, a package only ever imported via multi-line syntax would score PROD_HITS=0 and be auto-dismissed as `not_used` — a false-negative dismissal of a package that actually ships to WASM. Fix: switch to `grep -Pzo` with a multiline pattern (e.g., `import[^;]*?from\s*['"]${PKG_RE}(/[^'"]*)?['"]`) or run AST-based detection via `tsc`/AS compiler output, and add a sanity probe (e.g., refuse to auto-dismiss `@graphprotocol/graph-ts` since it's structurally the one package that DOES ship).

.claude/skills/triage-security/SKILL.md
```diff
@@ -0,0 +1,976 @@
...
+- **§2.5.8** PoC harness explicitly out-of-scope. If a `moderate` / `low` dismissal is challenged by the maintainer, PoC verification is the documented escalation path.
+
+### 2.6 Transitive reverse-resolve
+
+For pkgs that appear in yarn.lock but nowhere else, figure out which **direct** dep pulled them in:
+
+```bash
+# Walk from the manifest's directory (subgraphs/<name>/ or root).
+MANIFEST_DIR=$(dirname "$MANIFEST_PATH")
+(cd "$MANIFEST_DIR" && yarn why "$PKG" 2>/dev/null | head -30) > "$TMP/yarn_why_${PKG}.txt"
```
DhairyaPatel7: Redirect path embeds the raw package name: `> "$TMP/yarn_why_${PKG}.txt"`. For scoped packages (e.g. `@graphprotocol/graph-ts`, `@types/node`) `$PKG` contains `/`, so the target becomes `$TMP/yarn_why_@graphprotocol/graph-ts.txt`. The parent `@graphprotocol` directory does not exist in $TMP, so bash redirection fails with `No such file or directory`; under the `set -euo pipefail` declared in Phase 0.0 the skill aborts mid-run, leaving any in-flight audit issue orphaned. The subsequent `grep ... "$TMP/yarn_why_${PKG}.txt"` on line 413 would also fail. Fix: sanitize the slash, e.g. `local safe_pkg="${PKG//\//_}"` and use `"$TMP/yarn_why_${safe_pkg}.txt"`, or `mkdir -p "$(dirname ...)"` before the redirect.

.claude/skills/triage-security/SKILL.md
```diff
@@ -0,0 +1,976 @@
...
+Step 2 — grep prod (AS source) vs dev (everything else):
+
+```bash
+# PROD — subgraphs/*/src/ AS sources. These are what actually ships to WASM.
+PROD_HIT_FILES=""
+if [[ ${#PROD_DIRS[@]} -gt 0 ]]; then
+  PROD_HIT_FILES=$(grep -rlnE "$IMPORT_RE" "${PROD_DIRS[@]}" \
+    --include="*.ts" --include="*.tsx" 2>/dev/null | grep -v "/generated/" | grep -v "/build/" || true)
+fi
+PROD_HITS=$(printf '%s\n' "$PROD_HIT_FILES" | grep -c . || echo 0)
```
DhairyaPatel7: `PROD_HITS=$(printf '%s\n' "$PROD_HIT_FILES" | grep -c . || echo 0)` produces a two-line value when PROD_HIT_FILES is empty: `printf '%s\n' ""` emits a single empty line; `grep -c .` finds zero matches, prints `0`, exits 1; the `|| echo 0` fires and prints another `0`. PROD_HITS ends up as the literal string `0\n0`, breaking any later `[[ $PROD_HITS -gt 0 ]]` arithmetic and producing surprising verdicts. Same bug for DEV_HITS at line 341. Fix: short-circuit on empty input, e.g. `if [[ -z "$PROD_HIT_FILES" ]]; then PROD_HITS=0; else PROD_HITS=$(printf '%s\n' "$PROD_HIT_FILES" | wc -l | tr -d ' '); fi`.

.claude/skills/triage-security/SKILL.md
```diff
@@ -0,0 +1,976 @@
...
+
+**Apply across every package.json** (root + 12 subgraphs — the repo enforces resolution parity per CLAUDE.md "Yarn 1 gotcha").
+
+For direct deps (only \`@graphprotocol/graph-cli\`, \`@graphprotocol/graph-ts\`, \`matchstick-as\` qualify): bump the dependency pin in every package.json that declares it.
+
+After any resolution change, **delete every yarn.lock** (root + all subgraphs) and run \`yarn install\` fresh. The lockfile is sticky; without wiping it, yarn won't re-resolve even when the resolution is correct. A fresh install also auto-bumps every transitive to its latest semver-compatible patch — often clearing several other CVEs in the same PR without explicit resolutions for them.
+
+\`\`\`bash
+# Apply the resolution edits to all 13 package.json (root + 12 subgraphs), then:
+rm yarn.lock subgraphs/*/yarn.lock subgraphs/*/*/yarn.lock
```
DhairyaPatel7: The fix recipe inside the issue body uses `rm yarn.lock subgraphs/*/yarn.lock subgraphs/*/*/yarn.lock` with no `-f`. If any of the three globs match no files (e.g. a fresh checkout where the nested subgraphs/*/*/ pattern has no yarn.lock), bash leaves the unexpanded literal pattern in place and rm errors out, aborting the cascade before `yarn install` runs. Use `rm -f yarn.lock subgraphs/*/yarn.lock subgraphs/*/*/yarn.lock` so a missing path is a no-op.

.claude/skills/triage-security/SKILL.md
```diff
@@ -0,0 +1,976 @@
...
+if [[ -n "$EXISTING" ]]; then
+  echo "skip: existing issue for $GHSA_ID at $EXISTING"
+  SKIPPED+=("$ALERT_NUM:existing-issue:$EXISTING")
+  continue
+fi
+
+# Pre-create labels idempotently. `gh issue create --label X` errors with
+# "could not add label: 'X' not found" if the label doesn't already exist.
+# Run once before the per-alert loop (NOT per alert):
+gh label create security             --color B60205 --description "Security vulnerability" --repo "$REPO" 2>/dev/null || true
```
DhairyaPatel7: The five `gh label create ... 2>/dev/null || true` calls sit inline inside §3.2 immediately above the per-alert `gh issue create`, with only a code comment (`# Run once before the per-alert loop (NOT per alert)`) marking them as setup. Anyone copy-pasting §3.2 verbatim into a loop will invoke label creation on every alert — five gh API calls per alert against a labels endpoint that doesn't dedupe at the wire level. The reference loop (line ~858) correctly hoists these out. Move the label block out of §3.2 into Phase 0 or a clearly fenced `## Setup` block, so the per-alert region of §3.2 starts with the dedupe and ends with the issue create.

.claude/skills/triage-security/SKILL.md
```diff
@@ -0,0 +1,976 @@
...
+
+**Re-evaluate if:** any AS file under \`subgraphs/*/src/\` adds an import of \`${PKG}\`, or the deployed-runtime model changes (e.g. graph-node starts accepting JS handlers).
+EOF
+)
+      ;;
+    inaccurate)
+      body_analysis=$(cat <<EOF
+## Advisory summary (§2.5.1)
+
+${ADVISORY_SUMMARY}
```
DhairyaPatel7, line 661: The `inaccurate` audit heredoc references `${ADVISORY_SUMMARY}`, `${CWE_CHECKLIST_ANSWERS}`, `${SYMBOL_TRACE_RESULT}`, `${APPLICABILITY_REASONING}` — none of these are explicitly assigned in the visible Phase 2.5 code (only described prose-style in §2.5.1-§2.5.5). Combined with the `set -u` enabled in Phase 0.0, any unset variable here crashes the skill at audit-issue-create time, after the dismissal API patch has already happened in §3.1b (line 514) — leaving the alert dismissed with no audit trail. Either set safe defaults (e.g. `: "${ADVISORY_SUMMARY:=}"`) before the heredoc, or document the contract that §2.5 MUST populate these as a hard precondition for entering 3.1b.

.claude/skills/triage-security/SKILL.md
```diff
@@ -0,0 +1,976 @@
...
+}
+```
+
+The `security-audit` label and the other four labels MUST be pre-created idempotently before the per-alert loop runs (see §3.2).
+
+### 3.2 Open a tracking issue for an APPLICABLE alert
+
+```bash
+# Dedupe by GHSA ID — search both title and body.
+EXISTING=$(gh issue list --repo "$REPO" --state open --search "\"$GHSA_ID\" in:title,body" --json number,url --jq '.[0].url // ""')
```
DhairyaPatel7: Dedupe uses `--state open` only. If a maintainer manually closes a previously-opened tracking issue (e.g. after fixing the underlying vuln, or because they judged the auto-opened issue redundant), and the underlying Dependabot alert is still open on the next skill run, the dedupe misses the closed issue and a fresh duplicate is opened. Either search `--state all`, or `gh issue list --state closed` as a second pass and skip-with-log when a closed match is found.

Tanya-atatakai (dismissed):
## Approve

Solid skill design — the audit-trail invariant (closed issue created *before* the dismissal PATCH, URL embedded in the 280-char comment), the strict allowlist deferral to `.supply-chain/audit-allowlist.json`, and the `--rerun-dismissed` read-only mode are all the right guardrails for something that mutates GitHub state.

### One bug worth fixing before merge

**Empty-string `grep -c` produces `"0\n0"`** under `set -euo pipefail` (§2.4, lines ~336 / ~347):

```bash
PROD_HITS=$(printf '%s\n' "$PROD_HIT_FILES" | grep -c . || echo 0)
```

When `$PROD_HIT_FILES` is empty, `printf` emits a blank line, `grep -c .` prints `0` *and* exits non-zero, then `|| echo 0` runs and appends another `0`. The captured value becomes `"0\n0"`, and `[[ "$PROD_HITS" -gt 0 ]]` will misbehave on the first DEV-only alert. Cheap fix: gate on `[[ -z "$PROD_HIT_FILES" ]]` first.

### Smaller items (non-blocking)

- **§2.4 `@types/*` short-circuit** sets `VERDICT=DEV` but doesn't `continue` — the inline `OPERATOR:` comment papers over this. Since `disable-model-invocation: true`, Claude will route correctly, but a literal `bash`-extraction would fall through into the import-grep blocks. Either add `continue` in the loop context or move the short-circuit into the reference loop body.
- **`SIBLING_PKG_JSON="${MANIFEST_PATH%yarn.lock}package.json"`** (§2.3) corrupts when `manifest_path` is already a `package.json` — produces `…/package.jsonpackage.json`. Downstream verdict is still correct (falls back to `transitive`), but the diagnostic is misleading.
- **Dedupe race** (§3.2): `gh issue list --search` lags ~minutes after creation; back-to-back runs could double-open. Low risk for single-invocation use.
- **Fixed `/tmp/ts_*` paths** trample on concurrent runs. `mktemp` would be safer.
- **PR body says "runs fully autonomously"** but the reference loop has Phase 2/3 as `# … stub …` comments. Aligning the language with "methodology Claude interprets" would be more accurate.

### Approves

- Layout / disable-model-invocation / argument-hint frontmatter all correct.
- Hard rules in §"Hard rules" + Files-mutated table form a clean contract — forbidden `tolerable_risk`/`fix_started`/`no_bandwidth` reasons, audit-issue-before-dismissal, dedupe-by-GHSA, conservative-when-uncertain.
- `**/parent/child` Yarn 1 finding in Design Notes is exactly the kind of non-obvious lesson worth embedding so future maintainers don't re-learn it from the (now-empty) audit allowlist.

Squash-and-merge as the PR body recommends. 🤖

.claude/skills/triage-security/SKILL.md
mohandast52: Fixed in c5981b2. Switched to a two-pass approach:
- Pass 1: `grep -Pzo` with the `import { ... } from "<pkg>"` pattern — `-z` treats the file as null-terminated so `.` spans newlines and the multi-line import in `balancerVault.ts` (and similar) is now matched.
- Pass 2: `grep -E` for inline `require()` / dynamic `import()`.

Plus the belt-and-braces sanity probe you implied: if `@graphprotocol/graph-ts` (the one package that DOES ship to WASM) ever shows PROD_HITS=0, the skill forces PROD verdict + WARN to stderr rather than auto-dismissing — defends against future regex-evolution bugs.

GNU grep dependency made explicit: probed at startup, errors with `brew install grep` hint on macOS BSD grep.

.claude/skills/triage-security/SKILL.md
mohandast52: Fixed in c5981b2 — `PKG_SAFE="${PKG//\//_}"` and the temp file path uses `${PKG_SAFE}` instead of `${PKG}`. Verified the read on the next line uses the same sanitized name.

.claude/skills/triage-security/SKILL.md
mohandast52: Fixed in c5981b2 for both PROD_HITS and DEV_HITS. Replaced the `grep -c . || echo 0` form with an explicit empty-check + `wc -l | tr -d` for the count:

```bash
if [[ -z "$PROD_HIT_FILES" ]]; then
  PROD_HITS=0
else
  PROD_HITS=$(printf '%s\n' "$PROD_HIT_FILES" | wc -l | tr -d ' ')
fi
```

No more `0\n0` byte-string masquerading as an integer.

.claude/skills/triage-security/SKILL.md
mohandast52: Fixed in c5981b2 — all `rm yarn.lock ...` occurrences (one in the §3.2 issue body recipe, one in the Design Notes section) now use `rm -f`. Used `replace_all` so future drift would also need to use `-f`.

.claude/skills/triage-security/SKILL.md
mohandast52: Fixed in c5981b2. Created a new **Phase 0.7 — Pre-create labels (idempotent setup)** subsection that holds all 5 `gh label create` calls, with an explicit note about why this MUST be setup-only (the 5-API-calls-per-alert hazard you flagged). The §3.2 prose now has a one-liner pointing back: `(Label pre-creation lives in §0.7 — Phase 0 setup. Do NOT inline it here.)`.

.claude/skills/triage-security/SKILL.md
mohandast52: Fixed in c5981b2. Added `: "${VAR:=(not populated — see §2.5.X)}"` defaults at the top of `create_audit_issue()` for all 13 vars the heredoc references (ADVISORY_SUMMARY, CWE_CHECKLIST_ANSWERS, SYMBOL_TRACE_RESULT, APPLICABILITY_REASONING, plus the metadata vars).

The defaults are explicit `(not populated — see §2.5.X)` markers, not empty strings — so if §2.5 ever fails to populate them, the audit issue body shows a visible "not populated" rather than silently empty fields.

Also documented the contract in a comment block above the defaults: §2.5 MUST populate these as a precondition for entering §3.1b; the defaults are a safety net, not the spec.

.claude/skills/triage-security/SKILL.md
mohandast52: Fixed in c5981b2 — dedupe is now `--state all`, capturing the matched issue's state and surfacing it in the skip log:

```bash
EXISTING=$(gh issue list --repo "$REPO" --state all --search "\"$GHSA_ID\" in:title,body" --json number,url,state --jq '.[0]')
if [[ -n "$EXISTING" ]]; then
  EXISTING_URL=$(echo "$EXISTING" | jq -r '.url')
  EXISTING_STATE=$(echo "$EXISTING" | jq -r '.state')
  echo "skip: existing $EXISTING_STATE issue for $GHSA_ID at $EXISTING_URL"
  SKIPPED+=("$ALERT_NUM:existing-issue($EXISTING_STATE):$EXISTING_URL")
  continue
fi
```

This way if a maintainer manually closed a prior tracking issue, the skill sees the closed match and skips rather than opening a duplicate.

rajat2502 (approved):

DhairyaPatel7 (approved):

Tanya-atatakai (approved):

PR #129, opened 2026-05-22
docs(pearl-funds): implementation plan for Pearl funds-movement subgraph

kupermind:
## Summary

Plan document only — **no code**. Opened for verification before implementation begins.

Scopes a new `subgraphs/pearl-funds/` subgraph that indexes **funds movement for the Master Safe and Agent Safe of Pearl predict services** on Gnosis and Polygon.

This is the dedicated funding subgraph explicitly deferred by `subgraphs/pearl/SUBGRAPH_PLAN.md` §6.1 (*"a dedicated funding subgraph is the right place — not mixed into trades"*). It is **orthogonal** to the `docs/pearl-subgraph-plan` branch's `predict-polymarket` trade-tracking generalization — complementary, not a duplicate. In-market bet P&L stays owned by the predict subgraphs; consumers join on the Agent Safe address.

## What the plan covers

- **Phase 1 — semantic ledger:** Master Safe ↔ Agent Safe ↔ Service ↔ staking-contract graph; service-NFT custody trail; staking stake/claim/unstake/eviction with exact OLAS reward amounts (straight from events, no token indexing). Low cost, ships first.
- **Phase 2 — raw token ledger:** *2a* — OLAS `Transfer` data source + native coin via `Safe` dynamic templates (cheap, unconditional). *2b* — USDC / USDC.e(pUSD) ledger, **benchmark-gated** on the Polygon USDC.e volume risk, with the off-chain (Dune) fallback documented.
- Full asset/file inventory, reuse map (`staking`, `service-registry`, `babydegen`, prior pearl docs), implementation sequence, testing strategy.
- Inherits the on-chain-data-only / no-server-side-joins constraint from the prior Pearl plan (§2.1, §12 "Deliberately Absent").

## Key decisions captured

- New subgraph, template pattern → 2 Studio deployments (Gnosis, Polygon). Not folded into `service-registry` / `staking` / `predict-polymarket`.
- Pearl predict cohort keyed on agent ID: **Gnosis 25**, **Polygon 86**.
- Raw token amounts only (no USD valuation).

## Notes for reviewers

- Open items are listed in §11 — most notably the Phase 2b USDC.e indexing-cost benchmark and the Polygon `ServiceRegistryL2` start block.
- `valory-xyz/autonolas-subgraph` PR #89 (confirms agent ID 25 for Omen) is **still open and unported** — flagged in §11 as separate follow-up work.

Document: `subgraphs/pearl-funds/IMPLEMENTATION-PLAN.md`

PR #130, opened 2026-05-26
scaffold(pearl-transactions): empty-but-valid subgraph + CI matrix entry

kupermind:
## Summary

Step 2 of [`IMPLEMENTATION-PLAN.md`](https://github.com/valory-xyz/autonolas-subgraph-studio/blob/docs/pearl-funds-plan/subgraphs/pearl-funds/IMPLEMENTATION-PLAN.md) §8 (plan PR #129). Lands an **empty-but-valid** `subgraphs/pearl-transactions/` so CI is green and follow-up phase PRs can land code in isolation without re-litigating infra.

Name `pearl-transactions` per plan §11 #5 / @Tanya-atatakai's PR #129 review recommendation.

## What's in this PR

- `subgraphs/pearl-transactions/` with the converged pinned toolchain (`graph-cli` 0.98.1, `graph-ts` 0.38.2, `matchstick-as` 0.6.0).
- Template-pattern manifest (`subgraph.template.yaml` + `networks.json` + `scripts/generate-manifests.js`) covering Gnosis, Polygon, Optimism, Base.
- One `ServiceRegistryL2.RegisterInstance` handler with a single-field placeholder `Service` entity — deliberate stub, replaced by Phase 1a.
- One Matchstick smoke test.
- `CLAUDE.md` and `README.md`.

## Repo-level integration

- `.github/workflows/ci.yml` — adds `pearl-transactions` to test matrix and lockfile-lint.
- Root `CLAUDE.md` — adds `pearl-transactions/` to subgraph tree.

## Product decisions confirmed (see comments)

- **AC #3 — "Setup complete"**: Path A confirmed. Frontend renders opening balances via archive RPC (`eth_getBalance` / `token.balanceOf`) at `MasterSafe.historyFloorBlock`. No `OPENING_BALANCE` rows emitted by the subgraph.
- **Native → Agent EOA**: confirmed accepted known gap for v1.

## Open Qs resolved

- **#1** Polygon start block — `41,783,952` ✅
- **#3** `ServiceRegistryL2.json` ABI has ERC-721 Transfer ✅
- **#5** Subgraph name — `pearl-transactions` ✅
- **#6** graph-node backdated template startBlock — NOT SUPPORTED upstream ✅
- **#8** Pre-Master-Safe EOA history — `historyFloorBlock` is the contract, frontend handles opening balances via RPC ✅

Open Qs #2 (Phase 2b benchmark) and #7 (SRTU start blocks) deferred to follow-up PRs.

## Test plan

- [x] `yarn install`, `yarn generate-manifests`, `yarn codegen`, `yarn test` all pass
- [x] audit, install-hooks, lockfile-lint, gitleaks all clean
- [ ] CI green

## Follow-up PRs

| Step | Adds |
|---|---|
| Phase 1a (#131) | ServiceRegistryTokenUtility + Master Safe/EOA derivation + bond rows |
| Phase 1b (#132) | StakingFactory + StakingProxy + reward/unstake rows |
| Phase 2a (#133) | OLAS + WrappedNative ERC-20 + Safe templates + AgentFundingEvent |
| Phase 2b | USDC/USDC.e — benchmark-gated |

Related: plan PR #129.


Tanya-atatakai:
## Design review: gaps vs. the Figma "all tokens, all movements" requirement

Flagging this on the scaffold PR so it's on record before the phase PRs land. The Figma spec expects **every token used in Pearl** and **every movement of the addresses a user holds in Pearl** to be shown. Reviewing the as-built design (`phase-2a` branch + `IMPLEMENTATION-PLAN.md`), there are two categories of gap that won't meet that bar as currently scoped.

### 1. Token coverage gaps

What's actually tracked today (`phase-2a`): **OLAS** (dedicated ERC-20 `Transfer` data source, all 4 chains) and **native coin** (xDAI/POL/ETH via the per-Safe `Safe` template). Plus OLAS bonds/rewards read off events. That's it.

| Token | Status | Reason | Impact on Figma |
|---|---|---|---|
| USDC, USDC.e / **pUSD** | Deferred — Phase 2b, **not shipped** | Polygon USDC.e is a top-volume ERC-20; a full `Transfer` data source is the cost hotspot, so it's benchmark-gated (plan §6.3) | Stablecoin funding/spend — a primary Pearl flow on Polygon — is **invisible** today |
| **WXDAI**, WPOL/WMATIC | **No data source at all** (by design) | Plan assumes native coin via the `Safe` template suffices; wrapped tokens listed "only for symbol/decimals metadata" (§4.3) | **This is an architectural assumption, not just a deferral.** WXDAI is a *distinct* ERC-20 — its transfers are **not** seen by the `Safe` template (`SafeReceived` only fires for native value) and have **no** data source. If agents touch WXDAI (common on Gnosis: DEX swaps, many predict settlements), those movements drop out entirely. Same for WPOL/WMATIC. |

The WXDAI gap is the one to resolve before treating "native via Safe template" as complete. It's a cheap fix (WXDAI volume ≪ USDC.e → a normal ERC-20 data source, no benchmark fight), but only if we confirm agents actually use the wrapped form. The pUSD/USDC.e gap is the known, expensive one behind the benchmark gate.

### 2. Missing pre-discovery movements (account-setup history)

A Master Safe is first discovered at the **service-NFT mint** (`handleServiceNftTransfer` → `getOrCreateMasterSafe`), i.e. at service creation. Two hard graph-node rules govern what survives before that point:

- Dynamic `Safe` templates are **not retroactive** — a template `create()`'d at block N only sees events from N onward.
- Native transfers **to an EOA emit no log** — unobservable to any subgraph; there is no `eth_getBalance` binding in mappings (the code says so).

The one recovery mechanism is the OLAS **eth_call baseline**: at first sighting, `OLAS.balanceOf(masterSafe)` is written as one `SAFE_SETUP_TRANSFER` row. That recovers OLAS *value resting in the Master Safe*, but not the individual hops, and nothing for native.

Mapping the typical onboarding flow (1: user has only Master EOA → 2: funds Master EOA with OLAS + native → 3: Master Safe created, funds moved in, dust left on EOA → 4: service created → 5: Agent Safe created + funded):

| Movement | Tracked? | Why |
|---|---|---|
| OLAS → Master **EOA** (step 2) | ❌ | EOA not yet tracked; no EOA baseline |
| native → Master **EOA** (step 2) | ❌ **forever** | Native-to-EOA emits no log — chain-level invisible |
| OLAS mEOA → **Master Safe** (step 3) | ⚠️ value only | Not a transfer row; recovered as the lump-sum `SAFE_SETUP_TRANSFER` baseline at step 4 |
| native mEOA → **Master Safe** (step 3) | ❌ | Safe template not live until step 4; no native baseline — **this native is lost** |
| dust left on mEOA | ❌ | Same as step 2 |
| service NFT mint (step 4) | ✅ discovery point | Spawns Master Safe template, emits OLAS baseline, inits `TokenBalance` |
| Master Safe → Agent Safe (step 5) | ✅ | Both tracked; OLAS exact, native-in exact, native-**out** approximate (relayer 0-value) |

**Consequence for the UI:** for a freshly onboarded user, the history effectively *begins* at the OLAS baseline. We can show "X OLAS was in the Master Safe when first seen," but **cannot** reconstruct "user funded the EOA with Y, moved Z to the safe, kept dust for gas," and the **initial native funding of the account is not shown at all**. This matches what the plan lists as out of scope ("pre-Master-Safe-creation Master EOA history — fundamentally unrecoverable from on-chain events").

### Asks before the phase PRs

1. **Confirm the token set the Figma actually needs** per chain (esp. whether WXDAI / WPOL and USDC.e/pUSD are required). If wrapped natives are in scope, add the WXDAI/WPOL ERC-20 data sources — and decide the USDC.e benchmark question explicitly rather than letting it slip.
2. **Decide how the UI represents the pre-discovery floor** — e.g. an explicit "history starts here" marker / opening-balance row, so users don't read the OLAS baseline as the literal first transaction and don't expect native account-setup movements that can't exist on-chain.
3. If full account-setup history (incl. native EOA funding) is a hard product requirement, that needs **trace/call-level indexing or an off-chain source** (Dune / archive RPC keyed on the addresses) — it cannot come from event-based subgraph indexing.

_(Reviewer analysis against `phase-2a` as-built + plan; not blocking this scaffold PR — raising now so token scope and the pre-discovery UX are settled before Phase 2 implementation.)_

rajat2502:
Adding the wallet-UI consumer perspective on top of @Tanya-atatakai's review. Two confirmations on the token-set question and a concrete suggestion for the pre-discovery UX.

### Token coverage — the Figma confirms WXDAI is in scope

[VLOP-73 Figma node 17758:73120](https://www.figma.com/file/HB1eVqbkaFRYs6NXbUmuQM/-%3E%20Pearl%20%26%20Agent%20Wallet?node-id=17758-73120) shows two multi-token rows that mix wrapped + native + ERC-20s on Gnosis:

- **"Withdraw to external wallet"** — `-3.25 USDC.e / -23.15 XDAI / -5,255.87 OLAS / -6.53 WXDAI` (4 tokens, one tx)
- **"Omenstrat withdrawal"** — `+182.42 OLAS / +6.25 XDAI / +3.25 WXDAI` (3 tokens, one tx, AGENT_TO_MASTER reward sweep)

So both **WXDAI** and **USDC.e** are first-class displayed tokens, not edge cases. Without them:
- the Gnosis predict-trader settlement-sweep row (`AGENT_TO_MASTER`) under-reports — design shows 3 tokens, subgraph would emit 2 (OLAS + xDAI). User reads "Omenstrat returned 182.42 OLAS" and the WXDAI silently vanishes from the wallet view.
- the Polystrat funding row (`MASTER_TO_AGENT`/`AgentFundingEvent`) on Polygon shows nothing meaningful — Polystrat's bet capital is USDC.e-denominated, so `Fund Polystrat: -5.00 USDC.e` simply doesn't render.

**WXDAI**: agreeing with @Tanya-atatakai — looks like a cheap add (volume is bounded by Pearl-relevant addresses via the `TrackedSafe` in-handler filter), and the design needs it.

**USDC.e**: this is the Phase 2b benchmark + product question (plan §6.3). If 2b ships off-chain (Dune/RPC), the wallet UI ends up merging two data sources client-side — feasible but doubles the Polygon-side complexity. Worth a product decision **before** Phase 2 code lands so consumers know which path to wire for.

### Pre-discovery UX — anchor wording matters more than the data

Agreeing the pre-Master-Safe-creation history is fundamentally unrecoverable from events. From the consumer side, the friction is less "where's my account-setup history" and more "the row labeled 'Setup complete' shows a number that has no obvious relationship to what I actually deposited."

Concretely with Option 2 today: a typical Pearl user funds with native xDAI first (for gas), then a small OLAS top-up via the deposit flow. The `SAFE_SETUP_TRANSFER` baseline emits a single OLAS row with whatever's resting in the Safe at NFT-mint block — often 0 or a fractional amount that doesn't match the user's mental model of "setup."

Two consumer-side mitigations worth scoping before the schema is final:

1. **Distinguish the anchor from the "Setup complete" copy.** VLOP-73 currently asks for "first history entry = 'Setup complete' event." If `SAFE_DEPLOYED` (the synthetic anchor with amount=0) and `SAFE_SETUP_TRANSFER` (the OLAS baseline) are surfaced as separate categories already, the UI can render `SAFE_DEPLOYED` as a literal "Activity tracked from here" / "History begins" marker — no fabricated amount, no implied claim about what was deposited. The OLAS baseline then renders as a normal "Opening OLAS balance" row beneath it. This is design-only on our side; just needs a sign-off.

2. **Expose `MasterSafe.firstSeenBlock` in the query response** (it's already in the schema). Lets the frontend link to "View earlier transactions on Etherscan" with the right block-range hint for the Master EOA — handles option (c) from @Tanya-atatakai's review at the wallet level rather than as a subgraph extension.

If full pre-discovery account-setup history is a hard product requirement, +1 on the trace/call-level or off-chain conclusion — would belong in Pearl Wallet, not in this subgraph.

### Asks summary, from the consumer side

1. **WXDAI is required by the Figma** — please add the ERC-20 data source for Gnosis (`0xe91d153e0b41518a2ce8dd3d7944fa863463a97d`) if the volume/cost shape allows. Without it, two of the design's seven row types under-report.
2. **USDC.e/pUSD decision is blocking for Polygon** — would appreciate a product call on §6.3 before Phase 2 code lands.
3. **`SAFE_DEPLOYED` as an explicit "history starts here" anchor** is the cleanest pre-discovery UX from the wallet side. No schema change needed beyond what's already there — just a sign-off that the consumer can repurpose its display copy this way.

kupermind:
Thanks @Tanya-atatakai — both gaps addressed across plan PR #129 (Rev. 4 design contract) and code PR #133 (`phase-2a/pearl-transactions` branch).

## Gap 1 — token coverage

- **WXDAI / WPOL / WETH promoted from metadata-only to a `WrappedNative` ERC-20 `Transfer` data source per chain.** Plan §4.3 + §4.5 updated; code lands a 5th data source in `subgraph.template.yaml` with the same generic handler as OLAS (Transfer + balanceOf via ERC20Detailed.json). Closes the "WXDAI movement drops entirely" hole.
- **USDC.e / pUSD** stays Phase 2b benchmark-gated per §6.3 — confirmed explicitly (your second sub-ask). The Polygon USDC.e cost concern is real; we want the throwaway-subgraph measurement before committing to a chain-wide `Transfer` source.

## Gap 2 — pre-discovery UX

Split the overloaded "baseline = setup transfer" concept into three distinct things per your ask for an explicit "history starts here" marker:

- **New `OPENING_BALANCE` `FundsCategory`** — synthetic snapshot rows at first sighting. One per tracked ERC-20 (OLAS + WrappedNative) with `amount = eth_call balanceOf`, plus one zero-amount native marker row (`token = null`) so the wallet UI has a deterministic anchor to attach "native pre-discovery balance unknown" to.
- **New `MasterSafe.historyFloorBlock` + `historyFloorTimestamp`** — the UI uses these to render the literal "History starts here" divider. Named distinctly from internal `firstSeen*` fields because this is a consumer contract.
- **`SAFE_SETUP_TRANSFER` narrowed** — preserved, now fires only for the first **live** Master EOA → Master Safe hop **after** the `OPENING_BALANCE` baseline. The baseline does NOT flip `setupTransferSeen`, so a freshly-discovered Safe still sees its first inbound classified correctly. Verified in a new dedicated test.

## Gap 3 follow-on — native EOA pre-creation reframed

Plan §11 #8 rewritten to recommend (a) "document the floor as the contract" as shipped in Rev. 4. Pearl product picks between (b) `eth_call balanceOf(masterEoa)` snapshot or (c) off-chain Dune/Etherscan join before the Phase 9 docs PR closes the UX loop. Option (d) call/trace handlers documented as not viable (conflicts with §2.2 cost discipline).

## Refs

- Plan Rev. 4: PR #129 commit `da374b8`.
- Phase 2a code: PR #133 commit `84513e3` (+508 / −86 across 13 files).
- 34/34 Matchstick tests passing locally (was 32 — added 2 for the new behaviors: WrappedNative routing + setupTransferSeen non-pre-emption).

Re-requesting your review.

kupermind:
Thanks @rajat2502 — auditing your three asks against the Rev. 4 changes already pushed:

**Ask 1 — WXDAI required by Figma.** ✅ Done in Rev. 4 (PR #133 commit `84513e3`). New `WrappedNative` ERC-20 `Transfer` data source on Gnosis (`0xe91d153e0b41518a2ce8dd3d7944fa863463a97d`), same generic handler as OLAS, same `TrackedAddress` in-handler filter. WPOL/WETH ship symmetrically on Polygon/Optimism/Base. The "Omenstrat withdrawal" row should now report all three constituents (OLAS + xDAI + WXDAI); the "Withdraw to external wallet" row will too once Phase 2b lands USDC.e (see Ask 2).

**Ask 2 — USDC.e decision blocking for Polygon.** Plan §6.3 tightened in `5e65c5e` with your consumer-side merge-cost framing: off-chain fallback doubles Polygon wiring complexity in the wallet client and changes the offline-cache story per-chain. Decision now explicitly tagged "decide before Phase 2 code lands" with three inputs (infra cost / product / consumer cost). Still a product call to make — pinging Pearl product is the unblock.

**Ask 3 — `SAFE_DEPLOYED` as the "history starts here" anchor.** Already aligned with Rev. 4's split:
- `SAFE_DEPLOYED` (amount=0) stays the anchor — wallet UI renders it as the "Activity tracked from here" / "History begins" marker exactly as you propose. No fabricated amount.
- The OLAS baseline you referenced is **renamed in Rev. 4 to `OPENING_BALANCE`** — a new dedicated `FundsCategory`. One row per tracked ERC-20 (OLAS + WrappedNative) with `amount = eth_call balanceOf` plus a zero-amount native marker row. UI renders these directly as "Opening OLAS balance" / "Opening WXDAI balance" / "Opening native balance (unknown)" — no copy gymnastics needed, the category name carries the semantics.
- Both `MasterSafe.firstSeenBlock` (internal provenance) and the new `MasterSafe.historyFloorBlock` (consumer contract) are queryable — pick either for the Etherscan deep-link. They're equal in practice; the naming split is just to make the consumer contract explicit.

This means `SAFE_SETUP_TRANSFER` is now narrowed to its true semantic — only fires for the first **live** Master EOA → Master Safe hop after the baseline, never pre-empted by it. Verified in a new test.

**Bonus observation from your Figma examples** — the multi-token rows ("Withdraw to external wallet" 4 tokens, "Omenstrat withdrawal" 3 tokens) need per-tx grouping in both directions, not just the `MASTER_TO_AGENT` direction `AgentFundingEvent` already handles. Added as an open Q for Rev. 5 in §6.5: either consumers `GROUP BY txHash` client-side (v1 default — no schema change) or we add a generic `TxBundle` entity if the wallet team prefers server-side grouping. Doesn't block Rev. 4.

## Refs

- Plan Rev. 4 commits: `da374b8` + `5e65c5e` (PR #129, branch `docs/pearl-funds-plan`).
- Code: `84513e3` (PR #133, branch `phase-2a/pearl-transactions`). 34/34 Matchstick tests passing locally.

Re-requesting your review.

rajat2502:
Self-correcting two points in my [earlier comment](https://github.com/valory-xyz/autonolas-subgraph-studio/pull/130#issuecomment-4554542001) after rechecking against the VLOP-73 spec verbatim. The token-coverage asks stand — those line up with the spec ("all token types relevant to that chain are shown (native + agent tokens + OLAS)" + the explicit USDC.e example) and the Rev. 4 commit on PR #133 already adopts them. The other two points needed correcting.

### Correction 1 — "Setup complete" is the AC, not a copy choice

I framed `SAFE_DEPLOYED` vs `SAFE_SETUP_TRANSFER` labelling as a "design sign-off." That understated it. VLOP-73 AC #3 is verbatim:

> *"First history entry after Safe creation is the **'Setup complete'** event"*

And the transactions list names it as a real transfer:

> *"Funds moved from MasterEOA (Safe Setup)"*

So the AC requires a transfer row with the actual MasterEOA → Master Safe funding amount and token, labelled "Setup complete." `OPENING_BALANCE` (Rev. 4 / Option 2) shows the post-funding balance snapshot at first sighting — not the transfer. As shipped, the AC is **only satisfiable via Option 1** (`Template.createWithContext` backdated `startBlock`, plan Open Q #6).

What this means practically:
- **Open Q #6 is on the critical path for VLOP-73 AC #3.** Verifying the graph-node `startBlock`-in-context capability and switching `SAFE_SETUP_TRANSFER` to Option 1 emit semantics is what gets the AC met.
- **If Option 1 isn't feasible**, the AC can't be satisfied from on-chain data and the wallet UX needs an explicit product call to relax it (e.g. accept "Setup complete" rendering only the OPENING_BALANCE rows). That's a spec change, not a consumer-side polish.

### Correction 2 — withdrawing the Etherscan deep-link suggestion

VLOP-73 explicitly says:

> *"History scope: **from Pearl install onwards**"*

The "View earlier transactions on Etherscan" deep-link via `MasterSafe.firstSeenBlock` was out of scope. Withdrawing the ask — no subgraph change needed for that.

### Standing asks (unchanged, all now adopted in Rev. 4 ✅)

- WXDAI as a Phase 2a data source on Gnosis — **landed in Rev. 4**
- WPOL / WETH on Polygon / Optimism / Base — **landed in Rev. 4**
- `OPENING_BALANCE` + `MasterSafe.historyFloor*` as an explicit anchor — **landed in Rev. 4**

### Outstanding

- **USDC.e / pUSD** for Polystrat funding on Polygon — still Phase 2b benchmark + product call (plan §6.3). Polystrat's funding is USDC.e-denominated; without 2b, the "Funded Polystrat" row doesn't appear on Polygon, which fails VLOP-73's explicit example.
- **Open Q #6 (Option 1 createWithContext)** — now upgraded from "polish" to "critical path for VLOP-73 AC #3" per Correction 1 above.

VLOP-73 AC #6 ("subgraph delay surfaces a visible refresh indicator") is on track via `_meta.hasIndexingErrors` — no subgraph ask there, just confirming.

Tanya-atatakai:
Thanks @kupermind — Rev. 4 lands the token side well. The `WrappedNative` data source closes the WXDAI/WPOL/WETH hole, which was the priority architectural gap. No further subgraph ask there.

On the two outstanding points, I think we can keep the subgraph as-is and resolve both at the product + FE level rather than adding subgraph complexity:

**Pre-discovery mEOA→Master Safe movements.** Agreed the actual hops aren't recoverable from events once we rule out trace/call handlers (rightly out per §2.2). I'd propose we **stop short of Option 1 (`createWithContext` backdated `startBlock`, Open Q #6)** and treat `OPENING_BALANCE` + `historyFloorBlock` as the contract. The "Setup complete" / opening-balance presentation then gets handled **FE-side** in the wallet UI — render `historyFloorBlock` as the explicit "history starts here" divider and the `OPENING_BALANCE` rows as opening balances, rather than reconstructing a literal transfer row in the subgraph.

The caveat: this *relaxes VLOP-73 AC #3*, which as written asks for a real "Funds moved from MasterEOA (Safe Setup)" transfer row. So before we close this it needs sign-off on two fronts:
- **@rajat2502 (FE):** confirm the wallet UI can carry this — `historyFloorBlock` divider + `OPENING_BALANCE` rows standing in for the "Setup complete" entry, without the literal transfer row. This is the option you already sketched, so I think we're aligned, but want it explicit.
- **PM:** confirm relaxing AC #3 to an opening-balance snapshot + FE framing is acceptable.

If both sign off, no subgraph change needed and Open Q #6 drops off the critical path. If AC #3 is held as-is, Option 1 verification comes back on the table.

**Native pre-discovery + USDC.e** — same shape, also product + FE calls: (a) whether native setup funding being absent (only the `token: null` marker) is acceptable for v1, and (b) the §6.3 USDC.e benchmark/scope decision, which @rajat2502 flagged as an explicit Figma token on Polygon.

Net: none of these need more subgraph work right now — they need a product + FE decision. Proposing we take all three to the PM and @rajat2502 as a batch, and I'm good to approve the subgraph direction on that basis.

kupermind:
Verified Open Q #6 since both your follow-ups depended on it. **Result: Path B is dead** — backdated `startBlock` for dynamic templates is not supported in graph-node, never has been, no PR coming.

## Verification (cited in plan §11 #6 + §6.2)

- `chain/ethereum/src/data_source.rs:132` in graph-node master hardcodes `start_block: creation_block` for spawned templates. [source](https://github.com/graphprotocol/graph-node/blob/master/chain/ethereum/src/data_source.rs)
- `runtime/wasm/src/host_exports.rs` `data_source_create` takes `creation_block` directly from the host (current processing block); never inspects WASM `context` for a `startBlock` key. The `DataSourceContext.setBigInt("startBlock", …)` trick that's been circulating is community lore, not real.
- Tracking issue [`graph-node#902`](https://github.com/graphprotocol/graph-node/issues/902) ("Historical Block Scanning For Dynamic Data Sources", April 2019) was stale-bot closed July 2025 without an implementing PR.
- Latest graph-node release is [v0.43.0 (April 2025)](https://github.com/graphprotocol/graph-node/releases/tag/v0.43.0); no v0.36-v0.43 release notes mention backdated template start blocks.
- Studio runs hosted graph-node → cannot have a feature that doesn't exist upstream.

Plan §11 #6 marked RESOLVED as not-supported. §6.2 updated.

## What this means for the decision tree

@rajat2502's "Open Q #6 is on the critical path for VLOP-73 AC #3" framing was right *up until* the verification — Path B (literal MasterEOA → MasterSafe transfer row via backdated startBlock) is what would have satisfied AC #3 as written, and it's gone.

@Tanya-atatakai's proposal (stop short of Option 1, render "Setup complete" FE-side from `historyFloorBlock` + `OPENING_BALANCE`) is now **the operative answer** — not just one of two paths.

The remaining call is the AC #3 relaxation sign-off:

- **Path A (recommended)** — relax AC #3 wording to "Setup complete = `historyFloorBlock` divider + `OPENING_BALANCE` row set, rendered FE-side". No subgraph change. Needs:
  - @rajat2502 (FE) confirms the wallet UI can carry this without a literal transfer row.
  - PM confirms relaxing the AC wording is acceptable.

- **Path C (fallback if AC #3 is held as written)** — three on-chain workarounds documented in §6.2, all heavier re-architectures than Rev. 4:
  - **C-1: Discovery + graft + static dataSources** (lowest pain). Discovery deployment records every Master Safe + its first-stake block; regenerate `subgraph.yaml` with each Safe as a static `dataSource` carrying its own historical `startBlock`; graft the re-deployment. Cost: manifest regen + redeploy each time the Safe set grows; graft keeps re-sync cheap.
  - **C-2: Chain-wide deferred classification.** Drop the `TrackedAddress` filter on OLAS / wrapped-native data sources; store every Transfer temporarily; resolve into `FundsMovement` when a Safe is first sighted. Large storage footprint. Native still unrecoverable.
  - **C-3: Substreams.** Canonical "global filter" solution; needs Studio tier supporting substreams + a substreams package.

Recommend C-1 if AC #3 is held as written — but Path A is the realistic answer.

## Other outstanding items (unchanged)

- **USDC.e §6.3** — still product call.
- **Native pre-discovery** — still product call (option (a) "document the floor as the contract" is shipped in Rev. 4; (b)/(c) need PM).

## Refs

- Verification + plan updates: PR #129 commits `d1005fd` + `391a2f1` (branch `docs/pearl-funds-plan`).
- Subgraph code (PR #133) unchanged — Rev. 4 stands as the contract for Path A. 34/34 Matchstick tests passing.

Proposing we take this to Pearl PM + @rajat2502 (FE) as a batch decision: AC #3 relaxation (Path A) + native pre-discovery scope + USDC.e §6.3 path. I can prep code changes for any of the Path C workarounds if needed — say the word.

kupermind:
## Verdict on the AC #3 decision (and what's blocking the rest of the stack)

Posting a top-level take since the threaded back-and-forth has stretched across two review rounds. **The remaining blockers aren't engineering — they're product decisions that determine whether the code already on PR #133 is the right code.**

### What's actually decided

- **Token coverage gap (Gap 1) — closed.** Wrapped natives (WXDAI / WPOL / WETH) are Phase 2a data sources as of Rev. 4. USDC.e is a separate decision (§6.3).
- **Path B (`Template.createWithContext` backdated startBlock) — dead.** Verified upstream-not-supported; see §11 #6 resolution. graph-node hardcodes `start_block: creation_block` and never has shipped backdated template start blocks.

### What this leaves

For VLOP-73 AC #3 ("Setup complete" entry), the design space collapses to two options:

| Path | What it costs | What it gives up |
|---|---|---|
| **A** — Relax AC #3 wording, render "Setup complete" FE-side from `historyFloorBlock` + `OPENING_BALANCE`. | No subgraph change. ~1-hour PM + FE conversation. | The literal "Funds moved from MasterEOA (Safe Setup)" transfer row; users see an opening-balance snapshot instead. |
| **C-1** — Re-architect Phase 2a as discovery+graft+static-`dataSources`. | Phase 2a rework (substantial — different shape from the current dynamic-template pattern). Operational overhead forever: regen + redeploy + graft on every new Master Safe. | The dynamic, hands-off operational model. |

C-1 isn't a tweak. It throws out the dynamic Safe-template pattern and replaces it with regenerate-and-graft-on-every-new-Safe — a real operational tax, forever.

### My recommendation: push for Path A

- AC #3's literal wording (a real transfer row) was written before anyone confirmed graph-node couldn't deliver it. It's a spec assumption, not a user need. The user need is "the wallet shows what happened during onboarding" — and `historyFloorBlock` + `OPENING_BALANCE` covers that cleanly, with different copy.
- C-1 buys one literal row at the cost of operational overhead forever. Long-term tax for a one-time UX nicety.
- @Tanya-atatakai already proposed Path A; @rajat2502's original suggestion was Path A-shaped. If FE confirms the wallet UI can carry this, the call is 1 hour, not a planning effort.

### Sequencing I'd run

1. **Merge #134 first** (`tmp` resolution bump). Unblocks the audit gate on every other PR in the stack — independent fix, low risk.
2. **Merge #130 → #131 → #132 in order** (scaffold, Phase 1a, Phase 1b). Uncontroversial; AC #3 decision doesn't touch them.
3. **Hold #133** (Phase 2a) until the AC #3 call. If Path A: merge as-is. If C-1: I redo Phase 2a from the discovery+graft pattern.
4. Phase 2b USDC.e benchmark + native-EOA scope decisions can batch with the AC #3 conversation.

### What I'd push the team on

Schedule the PM + FE conversation explicitly — don't let it drift. **Three batched decisions** in one meeting unblock the entire stack:

1. AC #3 wording: literal transfer row (forces C-1 rework) or opening-balance snapshot (ships Rev. 4)?
2. Native pre-discovery: accept the floor as documented in §6.2, or pursue off-chain integration (§11 #8 option c)?
3. USDC.e / pUSD on Polygon: on-chain Phase 2b (benchmark first) or off-chain in wallet client (doubles client-side complexity per §6.3 Rev. 4)?

### Process win worth keeping

The plan-first approach + design review on the scaffold PR caught the WXDAI gap before we were weeks into broken indexing. That's the win that justifies the review-round overhead. Worth doing on future subgraphs.

rajat2502:
## Product decisions — unblocking the stack

Following internal review of the full options landscape (event-based subgraph, Substreams, Goldsky Mirror, self-hosted trace indexer), the product position is now confirmed:

**AC #3 — "Setup complete" = actual transfer row**
Accepted as-is with the Rev. 4 design. The `OPENING_BALANCE` snapshot row (OLAS + WrappedNative `balanceOf` at first sighting, plus the zero-amount native marker) is the product-agreed representation of the history floor. `SAFE_DEPLOYED` remains the UI anchor ("Activity tracked from here"). No further subgraph change needed.

**Native → Agent EOA (the 2 xDAI gas leg)**
Confirmed as a known, accepted limitation of event-based subgraph indexing. Native coin transfers to a plain EOA emit no log and are not observable from any event-based or call-handler-based subgraph on Gnosis/Polygon/Optimism/Base. This gap is documented and accepted for v1.

Both decisions unblock PRs #131 → #132 → #133. Requesting review on this scaffold PR so the stack can move forward.

rajat2502:
## Correction to previous comment + Rev. 5 update

The [previous comment](https://github.com/valory-xyz/autonolas-subgraph-studio/pull/130#issuecomment-4577022119) described `OPENING_BALANCE` snapshot rows as the product-agreed representation. This has since been revised in plan Rev. 5.

**Revised AC #3 position (Rev. 5):**
`OPENING_BALANCE` rows have been **removed from the subgraph** entirely. Opening balances are now the frontend's responsibility — the UI calls `eth_getBalance` and `token.balanceOf` against an archive RPC at `MasterSafe.historyFloorBlock` to derive them. The subgraph emits `historyFloorBlock` / `historyFloorTimestamp` as the anchor; everything else is frontend-side.

This means:
- No `OPENING_BALANCE` `FundsCategory` enum value in the final schema
- No `eth_call balanceOf` at first sighting in the handlers
- `SAFE_SETUP_TRANSFER` fires on the first live inbound hop directly (no longer gated behind an `OPENING_BALANCE` baseline)
- `setupTransferSeen` flag logic simplifies accordingly

**Native → Agent EOA gap** — unchanged, still accepted for v1.

Plan PR #129 updated to Rev. 5 with these decisions recorded.

Tanya-atatakai (approved):

rajat2502 (approved):
Approving — product decisions confirmed in comments (AC #3 Path A, native EOA gap accepted, plan updated to Rev. 5). CI green, Tanya approved. Good to merge.

PR #131, opened 2026-05-27
phase-1a(pearl-transactions): registry + SRTU bonds + Master EOA derivation

kupermind:
## Summary

Step 3 of [`IMPLEMENTATION-PLAN.md`](https://github.com/valory-xyz/autonolas-subgraph-studio/blob/main/subgraphs/pearl-funds/IMPLEMENTATION-PLAN.md) §8. Rebased on `main` (scaffold #130 merged; diff shows only Phase 1a changes).

Full semantic ledger for the registry + SRTU side of the subgraph: new entity types + enums, 8 event handlers across 2 data sources, Master EOA derivation via one-shot eth_call, a per-tx bond-type attribution queue, and Matchstick tests covering every documented behavior in plan §10's Phase 1 list.

## What landed

**Schema (`schema.graphql`):** `MasterSafe` (with `masterEoa` / `owners` / `threshold`), `AgentSafe`, `Service` (expanded with `agentIds`/`operators` + state + nftCustodian), `FundsMovement` (`immutable: false` — see attribution note below), `ServiceNftCustodyChange`, plus internal helpers (`ServiceIndex`, `PendingRegistration`, `PendingBondCounter`, `PendingBondRow`, `AgentBondAttributionGuard`).

**Handlers:**
- `src/service-registry.ts` — `handleRegisterInstance` / `handleActivateRegistration` / `handleCreateMultisigWithAgents` / `handleServiceNftTransfer` / `handleTerminateService` / `handleOperatorUnbond`.
- `src/service-registry-token-utility.ts` — `handleTokenDeposit` / `handleTokenRefund`.

**Master EOA derivation (`src/utils.ts` `getOrCreateMasterSafe`):** one-shot `GnosisSafe.getOwners()` + `getThreshold()` eth_call at first sighting. Idempotent. Emits a single `SAFE_DEPLOYED` synthetic `FundsMovement` row anchoring the wallet UI's "Setup complete" entry (ID `"safe-deployed:" + safeAddr`, never collides with log-derived IDs). The NFT recipient is **only** treated as a Master Safe when `getOwners()` resolves — non-Safe recipients (staking proxy on stake, EOAs, …) return `null` and are skipped, so they don't create a phantom `MasterSafe`/`SAFE_DEPLOYED` row or clobber the real link.

**Bond-type attribution queue (`src/utils.ts`):** per-tx FIFO. On-chain the SRTU event (`TokenDeposit`/`TokenRefund`) always fires **before** its `ServiceRegistryL2` counterpart (`ServiceManager` calls the `*TokenDeposit`/`*TokenRefund` function before the registry function in all four paths), so the SRTU handler is the **producer** — it creates the `FundsMovement` row and enqueues its id — and the `ServiceRegistryL2` handler (`ActivateRegistration`/`RegisterInstance`/`TerminateService`/`OperatorUnbond`) is the **consumer**, dequeuing the oldest pending row and backfilling `serviceId` + `bondType`. `RegisterInstance` (fires once per agent instance) dedupes via `AgentBondAttributionGuard` so the single agent-bond row is attributed exactly once per service per tx. Empty-queue case stays `null` per plan §5.4's best-effort limit.

**Data sources / addresses:**
- `subgraph.template.yaml` expanded with the 5 needed ServiceRegistryL2 events + new `ServiceRegistryTokenUtility` data source.
- `networks.json` — SRTU addresses + per-chain start blocks (resolves plan Open Q #7).
- `abis/ServiceRegistryTokenUtility.json` — sourced from `valory-xyz/autonolas-registries` at commit `ccead7b3`, normalized to plain ABI array.

## Review fixes (vs. first push)

Reviewed against the actual [autonolas-registries](https://github.com/valory-xyz/autonolas-registries) contracts; two correctness bugs fixed (details in [this comment](https://github.com/valory-xyz/autonolas-subgraph-studio/pull/131#issuecomment-4578043179)):

1. **Attribution queue direction was inverted.** `ServiceManager` emits the SRTU event before the SR event, so the original stash-on-SR / consume-on-SRTU design produced `null` or mislabeled `bondType` on real data. Roles swapped (producer = SRTU, consumer = SR); `FundsMovement` → `immutable: false` to allow the backfill.
2. **NFT transfer to a non-Safe corrupted `service.masterSafe`.** `StakingBase._stake` moves the NFT Master Safe → staking proxy; the old handler relabeled the service as owned by the proxy. Now gated on a `getOwners()` probe.

## Open Qs from PR #129 resolved here

- **#3** ABI verification — `ServiceRegistryL2.json` has all 5 needed events.
- **#7** SRTU start blocks — Gnosis 30,095,874 / Polygon 52,737,296 / Optimism 116,423,237 / Base 10,827,475 (all RPC-verified).

## Test plan

- [x] `yarn install` + `generate-manifests` + `codegen` + `test` — **10/10 Matchstick passing** (tests rewritten to the real on-chain event order; added staking-proxy skip + exactly-once agent-bond attribution cases).
- [x] `node scripts/audit-install-hooks.mjs` — OK (no root tree changes).
- [x] CI: `test / pearl-transactions` + root `Dependency audit` green on the rebased branch.

## Follow-up PRs (per plan §8)

- **Phase 1b** (step 4) — StakingFactory + StakingProxy template; reward / unstake / evict rows; DailyServiceFunds. Replaces the Phase-1a `getOwners()` probe with an explicit `StakingContract` allowlist for NFT-recipient classification.
- **Studio verification** (step 5).
- **Phase 2a** (step 7) / **Phase 2b** (step 8).
- Hygiene: bump `subgraphs/pearl-transactions` `tmp` resolution to `^0.2.6` for convergence once #134 lands (root audit already passes; this is convergence-policy only).

Related: plan PR #129, scaffold PR #130.

### DIFF
--- a/abis/ServiceRegistryTokenUtility.json
+++ b/abis/ServiceRegistryTokenUtility.json
@@ -1,0 +1,917 @@
+[
+  {
+    "inputs": [
+      {
+        "internalType": "address",
+        "name": "_serviceRegistry",
+        "type": "address"
+      }
+    ],
+    "stateMutability": "nonpayable",
+    "type": "constructor"
+  },
+  {
+    "inputs": [
+      {
+        "internalType": "address",
+        "name": "operator",
+        "type": "address"
+      }
+    ],
+    "name": "AgentInstanceRegistered",
+    "type": "error"
+  },
+  {
+    "inputs": [
+      {
+        "internalType": "uint256",
+        "name": "serviceId",
+        "type": "uint256"
+      }
+    ],
+    "name": "AgentInstancesSlotsFilled",
+    "type": "error"
+  },
+  {
+    "inputs": [
+      {
+        "internalType": "uint256",
+        "name": "agentId",
+        "type": "uint256"
+      }
+    ],
+    "name": "AgentNotFound",
+    "type": "error"
+  },
+  {
+    "inputs": [
+      {
+        "internalType": "uint256",
+        "name": "agentId",
+        "type": "uint256"
+      },
+      {
+        "internalType": "uint256",
+        "name": "serviceId",
+        "type": "uint256"
+      }
+    ],
+    "name": "AgentNotInService",
+    "type": "error"
+  },
+  {
+    "inputs": [
+      {
+        "internalType": "uint256",
+        "name": "componentId",
+        "type": "uint256"
+      }
+    ],
+    "name": "ComponentNotFound",
+    "type": "error"
+  },
+  {
+    "inputs": [],
+    "name": "HashExists",
+    "type": "error"
+  },
+  {
+    "inputs": [
+      {
+        "internalType": "uint256",
+        "name": "sent",
+        "type": "uint256"
+      },
+      {
+        "internalType": "uint256",
+        "name": "expected",
+        "type": "uint256"
+      },
+      {
+        "internalType": "uint256",
+        "name": "serviceId",
+        "type": "uint256"
+      }
+    ],
+    "name": "IncorrectAgentBondingValue",
+    "type": "error"
+  },
+  {
+    "inputs": [
+      {
+        "internalType": "uint256",
+        "name": "sent",
+        "type": "uint256"
+      },
+      {
+        "internalType": "uint256",
+        "name": "expected",
+        "type": "uint256"
+      },
+      {
+        "internalType": "uint256",
+        "name": "serviceId",
+        "type": "uint256"
+      }
+    ],
+    "name": "IncorrectRegistrationDepositValue",
+    "type": "error"
+  },
+  {
+    "inputs": [
+      {
+        "internalType": "address",
+        "name": "sender",
+        "type": "address"
+      },
+      {
+        "internalType": "address",
+        "name": "manager",
+        "type": "address"
+      }
+    ],
+    "name": "ManagerOnly",
+    "type": "error"
+  },
+  {
+    "inputs": [
+      {
+        "internalType": "address",
+        "name": "provided",
+        "type": "address"
+      },
+      {
+        "internalType": "address",
+        "name": "expected",
+        "type": "address"
+      },
+      {
+        "internalType": "uint256",
+        "name": "serviceId",
+        "type": "uint256"
+      }
+    ],
+    "name": "OnlyOwnServiceMultisig",
+    "type": "error"
+  },
+  {
+    "inputs": [
+      {
+        "internalType": "address",
+        "name": "operator",
+        "type": "address"
+      },
+      {
+        "internalType": "uint256",
+        "name": "serviceId",
+        "type": "uint256"
+      }
+    ],
+    "name": "OperatorHasNoInstances",
+    "type": "error"
+  },
+  {
+    "inputs": [
+      {
+        "internalType": "uint256",
+        "name": "provided",
+        "type": "uint256"
+      },
+      {
+        "internalType": "uint256",
+        "name": "max",
+        "type": "uint256"
+      }
+    ],
+    "name": "Overflow",
+    "type": "error"
+  },
+  {
+    "inputs": [
+      {
+        "internalType": "address",
+        "name": "sender",
+        "type": "address"
+      },
+      {
+        "internalType": "address",
+        "name": "owner",
+        "type": "address"
+      }
+    ],
+    "name": "OwnerOnly",
+    "type": "error"
+  },
+  {
+    "inputs": [],
+    "name": "Paused",
+    "type": "error"
+  },
+  {
+    "inputs": [],
+    "name": "ReentrancyGuard",
+    "type": "error"
+  },
+  {
+    "inputs": [
+      {
+        "internalType": "uint256",
+        "name": "serviceId",
+        "type": "uint256"
+      }
+    ],
+    "name": "ServiceMustBeInactive",
+    "type": "error"
+  },
+  {
+    "inputs": [
+      {
+        "internalType": "address",
+        "name": "token",
+        "type": "address"
+      }
+    ],
+    "name": "TokenRejected",
+    "type": "error"
+  },
+  {
+    "inputs": [
+      {
+        "internalType": "address",
+        "name": "token",
+        "type": "address"
+      },
+      {
+        "internalType": "address",
+        "name": "from",
+        "type": "address"
+      },
+      {
+        "internalType": "address",
+        "name": "to",
+        "type": "address"
+      },
+      {
+        "internalType": "uint256",
+        "name": "value",
+        "type": "uint256"
+      }
+    ],
+    "name": "TransferFailed",
+    "type": "error"
+  },
+  {
+    "inputs": [
+      {
+        "internalType": "address",
+        "name": "multisig",
+        "type": "address"
+      }
+    ],
+    "name": "UnauthorizedMultisig",
+    "type": "error"
+  },
+  {
+    "inputs": [
+      {
+        "internalType": "uint256",
+        "name": "agentId",
+        "type": "uint256"
+      }
+    ],
+    "name": "WrongAgentId",
+    "type": "error"
+  },
+  {
+    "inputs": [
+      {
+        "internalType": "uint256",
+        "name": "numValues1",
+        "type": "uint256"
+      },
+      {
+        "internalType": "uint256",
+        "name": "numValues2",
+        "type": "uint256"
+      }
+    ],
+    "name": "WrongArrayLength",
+    "type": "error"
+  },
+  {
+    "inputs": [
+      {
+        "internalType": "uint256",
+        "name": "serviceId",
+        "type": "uint256"
+      }
+    ],
+    "name": "WrongOperator",
+    "type": "error"
+  },
+  {
+    "inputs": [
+      {
+        "internalType": "uint256",
+        "name": "state",
+        "type": "uint256"
+      },
+      {
+        "internalType": "uint256",
+        "name": "serviceId",
+        "type": "uint256"
+      }
+    ],
+    "name": "WrongServiceState",
+    "type": "error"
+  },
+  {
+    "inputs": [
+      {
+        "internalType": "uint256",
+        "name": "currentThreshold",
+        "type": "uint256"
+      },
+      {
+        "internalType": "uint256",
+        "name": "minThreshold",
+        "type": "uint256"
+      },
+      {
+        "internalType": "uint256",
+        "name": "maxThreshold",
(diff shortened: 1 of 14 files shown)


rajat2502:
Rebased on `main` (scaffold commit auto-dropped) and pushed two correctness fixes found while reviewing against the actual [autonolas-registries](https://github.com/valory-xyz/autonolas-registries) contracts.

### 1. Bond-attribution queue direction was inverted
`ServiceManager` calls the SRTU `*TokenDeposit`/`*TokenRefund` function **before** the `ServiceRegistryL2` registry function in *every* path:

| Manager fn | first → emits | second → emits |
|---|---|---|
| `activateRegistration` | `activateRegistrationTokenDeposit` → **`TokenDeposit`** | `activateRegistration` → **`ActivateRegistration`** |
| `registerAgents` | `registerAgentsTokenDeposit` → **`TokenDeposit`** | `registerAgents` → **`RegisterInstance`** |
| `terminate` | `terminateTokenRefund` → **`TokenRefund`** | `terminate` → **`TerminateService`** |
| `unbond` | `unbondTokenRefund` → **`TokenRefund`** | `unbond` → **`OperatorUnbond`** |

So the SRTU event always fires first. The original design stashed on the SR event and consumed on the SRTU event (the reverse), so on real chain data the security-deposit row got `null` bondType and the agent-bond row was mislabeled `SECURITY_DEPOSIT` (batched tx), or both were `null` (separate txs). The old tests passed only because they fabricated the reversed order.

**Fix:** SRTU handler is now the **producer** (creates the `FundsMovement` row + enqueues its id); SR handler is the **consumer** (dequeues oldest pending row, backfills `serviceId` + `bondType`). Requires `FundsMovement` → `@entity(immutable: false)`. Renamed `PendingBondAttribution` → `PendingBondRow`, `AgentBondStashGuard` → `AgentBondAttributionGuard`, counter slots → `nextEnqueueSlot`/`nextDequeueSlot`.

### 2. NFT transfer to a non-Safe recipient corrupted `service.masterSafe`
`StakingBase._stake` does `safeTransferFrom(msg.sender, address(this), serviceId)` — on stake the service NFT moves Master Safe → staking proxy. The old handler treated any non-zero recipient as a Master Safe: it created a phantom `MasterSafe` + bogus `SAFE_DEPLOYED` row for the proxy and **overwrote** the real `service.masterSafe` link, so a staked service dropped out of any "movements for this Master Safe" query.

**Fix:** `getOrCreateMasterSafe` now probes `getOwners()` and returns `null` on revert/empty; the caller only links when it resolves. A stake hop no longer clobbers the link; unstake re-links when the NFT returns to the Safe. This is the Phase-1a heuristic the plan/CLAUDE.md deferred to the 1b `StakingContract` allowlist — but available now without that entity set.

Tests rewritten to the real on-chain event order, plus new cases for the staking-proxy skip and exactly-once agent-bond attribution. **10/10 Matchstick passing**, codegen clean.

kupermind:
Reviewed your fix against the upstream `autonolas-registries` contracts (commit reachable from `main`). Both root-cause claims **independently confirmed**, and the test rewrite cleanly mirrors real on-chain ordering. CI is all-green ✅.

## Verification — Claim 1 (SRTU fires before SR in every path)

Confirmed in `contracts/ServiceManager.sol`. Manager wraps each call with the SRTU call first, registry second:

| Manager fn | SRTU call (first) | Registry call (second) |
|---|---|---|
| `activateRegistration` (L336–362) | L345 → `TokenDeposit` | L355/358 → `ActivateRegistration` |
| `registerAgents` (L369–413)       | L388–389 → `TokenDeposit` | L402/407 → `RegisterInstance` |
| `terminate` (L456–475)            | L464 → `TokenRefund`    | L467 → `TerminateService` |
| `unbond` (L481–500)               | L489 → `TokenRefund`    | L492 → `OperatorUnbond` |

Access-control sanity-check: `ServiceRegistryL2` gates all four entrypoints on `msg.sender == manager`, so an operator/EOA can't bypass the manager and emit SR events without a prior SRTU event. The producer/consumer flip is the right design.

## Verification — Claim 2 (NFT moves Master Safe → staking proxy on stake)

Confirmed: `contracts/staking/StakingBase.sol`
- `_stake` L865: `IService(serviceRegistry).safeTransferFrom(msg.sender, address(this), serviceId);` — NFT to `address(this)` (staking proxy)
- `_unstake` L946: `IService(serviceRegistry).transferFrom(address(this), msg.sender, serviceId);` — NFT back to original staker (`sInfo.owner` from L831)

The `getOwners()` revert-probe in `getOrCreateMasterSafe` is the right interim guard; the cited Phase 1b `StakingContract`-allowlist follow-up will tighten it further.

## Test changes — LGTM

- Stake/unstake-cycle tests reordered to real SRTU-first sequence.
- New `STAKING_PROXY` test confirms the link-preservation: `getOwners()` reverts → no phantom Master Safe, no `SAFE_DEPLOYED`, `service.masterSafe` keeps the mint-time link, `nftCustodian` still tracks the hop.
- New "two RegisterInstance + one TokenDeposit" test pins the dedupe-guard contract.
- Per-tx isolation test reframed correctly: depositing in txA, dequeuing in txB no-ops; txA's own dequeue still attributes. Stronger than the prior version.

## Two upstream edge cases worth noting (no code change requested)

Surfaced while verifying — current design handles them safely, but they deserve a one-liner in the limitations section:

1. **ETH-secured services emit no SRTU events at all.** Every SRTU emit sits inside `if (token != address(0))` (SRTU L368/L424/L489/L522). For native-secured services only the registry event fires, so `dequeueAndAttribute` finds an empty queue and silently no-ops. Net result: no `SERVICE_BOND_DEPOSIT` / `SERVICE_BOND_REFUND` rows for ETH-secured services. Correct (the amount lives in the registry contract, not SRTU), but consumers should know that bond rows are token-secured-only.

2. **`unbondTokenRefund` additionally guards on `refund > 0` (SRTU L533).** A slashed-to-zero operator emits no `TokenRefund` but still emits `OperatorUnbond`. In an isolated unbond tx the queue is empty → no-op, fine. The (very unusual) hazard would be a batched tx that unbonds multiple operators where one is fully slashed: the queue could shift by one. Not blocking, but worth mentioning if/when Phase 1b touches this path.

3. **`unbondWithSignature` (SRTU L514–560)** follows the same SRTU-before-SR ordering, so the producer/consumer queue covers it without changes. Not tested explicitly but no extra coverage needed.

Suggest adding a single bullet under "Honest limits" in `subgraphs/pearl-transactions/CLAUDE.md`:

> - SRTU bond rows are token-secured-only: ETH-secured services emit no `TokenDeposit`/`TokenRefund`, so no `SERVICE_BOND_*` rows are produced for them.

Otherwise: thanks for the catch — both bugs would have shipped silently and only surfaced once we cross-checked against real Pearl tx data. Approving in spirit; ready to merge after #134 lands and the audit gate clears.

rajat2502:
Thanks for the thorough re-verification @kupermind 🙏

Added both edge cases to **Honest limits** in `subgraphs/pearl-transactions/CLAUDE.md` (commit 8b98f8b):
- **Token-secured-only bond rows** — your exact suggested bullet: ETH/native-secured services emit no `TokenDeposit`/`TokenRefund` (guard `token != address(0)`), so no `SERVICE_BOND_*` rows; `dequeueAndAttribute` no-ops on the empty queue.
- **`unbondTokenRefund` `refund > 0` guard** — fully-slashed operator emits `OperatorUnbond` without `TokenRefund`; harmless in isolated unbond, the only hazard being a batched multi-operator unbond with one fully-slashed (not in Pearl's one-operator-per-service flow). Flagged to revisit if Phase 1b+ touches batched unbonds.

Agree `unbondWithSignature` needs no extra coverage (same SRTU-before-SR ordering).

One correction on merge-readiness: **#134 isn't a blocker for this PR.** After the rebase onto `main`, the root `Dependency audit (root tree)` check is already **green** here (the `tmp <0.2.6` advisory only affected the pre-rebase branch; `main`'s resolution set clears it). So this is mergeable on the audit front today — #134 is independent supply-chain hygiene. The only remaining gate is the review approval.

rajat2502 (approved):

DIvyaNautiyal07 (approved):
Approved. The producer/consumer ordering fix and the non-Safe NFT recipient gating both look right, and the test coverage for the on-chain event order is solid.

A few nice-to-haves for this PR or a follow-up:

- `subgraphs/pearl-transactions/CLAUDE.md` still references the old entity names `PendingBondAttribution` and `AgentBondStashGuard` (lines 24-25, 28) and says "8 Matchstick tests" (line 56), but they were renamed to `PendingBondRow` / `AgentBondAttributionGuard` in the fix commit and the file now has 10 tests.

- `assertBondRow` in `tests/service-registry.test.ts` doesn't check `from` / `to`. The deposit handler sets `from = account, to = SRTU` and the refund handler inverts them, so swapping them in either handler would pass every test today.

- `src/utils.ts:123-126` — when `getThreshold()` reverts, the code silently stores `threshold = 0`, which is not a valid Safe configuration. The `getOwners` revert path logs and skips, so the `getThreshold` path is asymmetric. Logging a warning (or treating it the same way) would keep the two consistent.

- The `AgentBondAttributionGuard` dedupe isn't directly exercised by the current "Two RegisterInstance + one agent-bond TokenDeposit" test (`tests/service-registry.test.ts:653-697`). Removing the guard early-return would still pass that test because the queue slot is already `attributed = true` by the time the second `RegisterInstance` fires. Adding a second prior unmatched `TokenDeposit` would make the dedupe failure visible.

- The empty-array branch of `getOwners()` (`src/utils.ts:100`, `ownersResult.value.length == 0`) is untested. Only the revert branch is exercised through the staking-proxy test.

- The internal helper entities (`PendingBondCounter`, `PendingBondRow`, `AgentBondAttributionGuard`, `PendingRegistration`) are never deleted after their purpose is served. Each enqueue/dequeue and each pending-drain leaves a permanent row. Calling `store.remove(...)` after attribution/drain would keep this bounded over years of indexing.

- `Service.state` is a freeform `String!` (`schema.graphql:51`) while `FundsCategory` and `ServiceBondType` are GraphQL enums. Same shape of constraint, but only the latter two are enforced at indexing time.

- `handleCreateMultisigWithAgents` in `src/service-registry.ts:92-96` calls `service.save()` twice — once after `drainPendingRegistration`, then again after assigning `agentSafe`. The first save is redundant.

- Test fixture `OPERATOR = MASTER_SAFE` (`tests/service-registry.test.ts:61`) uses the same address for both identities. A handler that wrote `event.params.operator` into the wrong field (e.g. `masterSafe` instead of `operators`) would slip past every test.

PR #132, opened 2026-05-27
phase-1b(pearl-transactions): StakingFactory + StakingProxy + reward/unstake/evict rows + DailyServiceFunds

kupermind:
## Summary

Step 4 of [`IMPLEMENTATION-PLAN.md`](https://github.com/valory-xyz/autonolas-subgraph-studio/blob/main/subgraphs/pearl-funds/IMPLEMENTATION-PLAN.md) §8. **Stacked on PR #131** (Phase 1a) — base branch is `phase-1a/pearl-transactions`. When #131 merges, GitHub retargets this to `main` and the diff narrows to just the Phase 1b additions.

Completes Phase 1 of the semantic ledger: full staking-side coverage (factory + dynamic proxy template + reward / unstake / eviction rows + daily snapshots) plus the NFT-handler narrowing that resolves the Phase 1a staking-proxy TODO.

## What landed

**Schema:** `StakingContract` entity, `Service.currentStakingContract` + `Service.totalOlasRewardsClaimed`, `DailyServiceFunds`, three new `FundsCategory` values + `FundsMovement.stakingContract` / `FundsMovement.epoch`.

**Manifest / data sources:** `StakingFactory` static + `StakingProxy` dynamic template (events: `ServiceStaked`, `RewardClaimed`, `ServiceUnstaked`, `ServiceForceUnstaked`, `ServicesEvicted`). `networks.json` extended with StakingFactory per-network addresses; `isAllowedImplementation` allow-list converged with `subgraphs/staking/src/utils.ts`.

**Handlers:**
- `src/staking-factory.ts` `handleInstanceCreated` — implementation allow-list guard; eth_call config snapshot; spawn template.
- `src/staking-proxy.ts` — 5 handlers: `handleServiceStaked` (canonical Master + Agent Safe discovery — `owner`/`multisig` are event params), `handleRewardClaimed` + daily rollup, `handleServiceUnstaked` / `handleServiceForceUnstaked`, `handleServicesEvicted` (one informational zero-amount row per affected service).
- `src/service-registry.ts` `handleServiceNftTransfer` — skips the staking-proxy custody hop.

## Rebase + review fixes (vs. first push)

Rebased onto the updated `phase-1a/pearl-transactions` tip (which carries the two Phase-1a review fixes). Two follow-ups here:

1. **NFT-recipient classification is now defence-in-depth.** The handler skips known staking proxies via `isStakingContract` (cheap, no eth_call) **and** falls through to the Phase-1a `getOwners()` null-probe for anything else. The probe is the safety net for proxies created before the `StakingFactory` start block (or by an older factory) that aren't in the `StakingContract` set — those would otherwise revert + clobber the link.
2. **`handleServiceStaked` null-checks `getOrCreateMasterSafe`** — Phase 1a made it return `MasterSafe | null`; `owner` is the Master Safe so it normally resolves, but the link is now guarded rather than assumed.

## Verification against `autonolas-registries` + `subgraphs/staking`

- `isAllowedImplementation` addresses match the canonical staking allow-list for all 4 networks (gnosis / matic / optimism / base).
- `StakingFactory` addresses + start blocks in `networks.json` match `subgraphs/staking/networks.json` exactly.
- `_stake` transfers the NFT Master Safe → staking proxy and records `sInfo.owner = msg.sender`, consistent with `ServiceStaked.owner` being the Master Safe; `_unstake` returns the NFT to the owner.

## Test plan

- [x] `yarn install` + `generate-manifests` + `codegen` + `test` — **17/17 Matchstick passing** (10 Phase-1a + 7 Phase-1b). `staking.test.ts` sets `dataSourceMock.setNetwork("gnosis")` (matchstick defaults to mainnet, which this subgraph's resolvers don't cover).
- [x] `node scripts/audit-install-hooks.mjs` — OK.
- [x] Root `Dependency audit` green on the rebased branch (the `tmp` advisory that affected the pre-rebase branch is no longer an issue post-rebase on `main`).

## Follow-up

| Step | Adds |
|---|---|
| 5 — Verify Phase 1 on Studio | spot-check stake/claim/unstake against block explorer |
| 6 — graph-node template `startBlock` verification | Open Q #6; gates §6.2 option |
| 7 — Phase 2a | OLAS + Safe templates + `AgentFundingEvent` + `SAFE_SETUP_TRANSFER` |
| 8 — Phase 2b | USDC / USDC.e benchmark + product decision |

### DIFF
--- a/subgraphs/pearl-transactions/networks.json
+++ b/subgraphs/pearl-transactions/networks.json
@@ -7,6 +7,10 @@
     "ServiceRegistryTokenUtility": {
       "address": "0xa45E64d13A30a51b91ae0eb182e88a40e9b18eD8",
       "startBlock": 30095874
+    },
+    "StakingFactory": {
+      "address": "0xb0228CA253A88Bc8eb4ca70BCAC8f87b381f4700",
+      "startBlock": 35206806
     }
   },
   "matic": {
@@ -17,6 +21,10 @@
     "ServiceRegistryTokenUtility": {
       "address": "0xa45E64d13A30a51b91ae0eb182e88a40e9b18eD8",
       "startBlock": 52737296
+    },
+    "StakingFactory": {
+      "address": "0x46C0D07F55d4F9B5Eed2Fc9680B5953e5fd7b461",
+      "startBlock": 62213142
     }
   },
   "optimism": {
@@ -27,6 +35,10 @@
     "ServiceRegistryTokenUtility": {
       "address": "0xBb7e1D6Cb6F243D6bdE81CE92a9f2aFF7Fbe7eac",
       "startBlock": 116423237
+    },
+    "StakingFactory": {
+      "address": "0xa45E64d13A30a51b91ae0eb182e88a40e9b18eD8",
+      "startBlock": 124618633
     }
   },
   "base": {
@@ -37,6 +49,10 @@
     "ServiceRegistryTokenUtility": {
       "address": "0x34C895f302D0b5cf52ec0Edd3945321EB0f83dd5",
       "startBlock": 10827475
+    },
+    "StakingFactory": {
+      "address": "0x1cEe30D08943EB58EFF84DD1AB44a6ee6FEff63a",
+      "startBlock": 17310019
     }
   }
 }
--- a/subgraphs/pearl-transactions/schema.graphql
+++ b/subgraphs/pearl-transactions/schema.graphql
@@ -1,19 +1,8 @@
-# pearl-transactions schema — Phase 1a.
+# pearl-transactions schema — Phase 1a + 1b.
 #
-# Matches IMPLEMENTATION-PLAN.md §5.1 (semantic ledger). Entities and
-# fields that only become populated in later phases are noted in inline
-# comments and added with placeholder cardinality so the schema does not
-# need to be rewritten for Phase 1b / 2 (additive evolution).
-#
-# Phase 1a deliberately omits the following from §5.1 / §6.5; they
-# arrive in their own PRs:
-#   - StakingContract entity + Service.currentStakingContract field
-#     (Phase 1b — StakingFactory + StakingProxy)
-#   - Service.totalOlasRewardsClaimed (Phase 1b)
-#   - DailyServiceFunds (Phase 1b)
-#   - AgentFundingEvent + FundsMovement.agentFundingEvent backref
-#     (Phase 2a)
-#   - Token / TrackedSafe / TrackedEOA / TokenBalance (Phase 2)
+# Matches IMPLEMENTATION-PLAN.md §5.1 (semantic ledger). Phase 2
+# entities (AgentFundingEvent, Token, TrackedSafe, TrackedEOA,
+# TokenBalance) arrive in their own PRs.
 
 # --- Structural -------------------------------------------------------
 
@@ -47,10 +36,25 @@ type Service @entity(immutable: false) {
   operators: [Bytes!]!                # deduplicated; sub-cohort filter
   masterSafe: MasterSafe
   agentSafe: AgentSafe
-  state: String!                      # REGISTERED|DEPLOYED|TERMINATED (STAKED|UNSTAKED added in Phase 1b)
+  state: String!                      # REGISTERED|DEPLOYED|STAKED|UNSTAKED|TERMINATED
   nftCustodian: Bytes                 # current ERC-721 owner
+  currentStakingContract: StakingContract  # set on stake, cleared on unstake
+  totalOlasRewardsClaimed: BigInt!    # cumulative across stake-cycles
   registeredTimestamp: BigInt!
   updatedTimestamp: BigInt!
+}
+
+# Staking proxy / "staking contract" — created on
+# StakingFactory.InstanceCreated for any allowed implementation. Config
+# snapshot is captured via contract calls at creation time; nothing
+# updates afterward.
+type StakingContract @entity(immutable: false) {
+  id: Bytes!                          # staking proxy address
+  implementation: Bytes!
+  minStakingDeposit: BigInt!
+  numAgentInstances: BigInt!
+  createdBlock: BigInt!
+  createdTimestamp: BigInt!
 }
 
 # --- Ledger -----------------------------------------------------------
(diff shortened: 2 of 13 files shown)


rajat2502:
Rebased onto the updated `phase-1a/pearl-transactions` (which now carries the two Phase-1a review fixes — inverted SRTU/SR attribution + non-Safe NFT skip). One conflict in `handleServiceNftTransfer`, resolved to keep **both** guards:

- `isStakingContract(to)` early-return (your Phase-1b explicit allowlist — cheap, no eth_call), **and**
- the Phase-1a `getOwners()` null-probe fall-through — the safety net for staking proxies created before the `StakingFactory` start block (or by an older factory) that aren't in the `StakingContract` set.

Also null-checked `getOrCreateMasterSafe` in `handleServiceStaked` (Phase 1a made it return `MasterSafe | null`; it would otherwise not compile).

Verified against `autonolas-registries` + `subgraphs/staking`:
- `isAllowedImplementation` matches the canonical allow-list for all 4 networks.
- `StakingFactory` addresses + start blocks match `subgraphs/staking/networks.json`.
- `_stake` does `safeTransferFrom(owner → proxy)` + sets `sInfo.owner = msg.sender`, consistent with `ServiceStaked.owner` = Master Safe.

17/17 Matchstick passing (10 Phase-1a + 7 Phase-1b); `test / pearl-transactions` + root audit green.

kupermind:
Reviewed your rebase + 8-line null-check against both `autonolas-registries` and `subgraphs/staking`. Everything checks out. CI all-green ✅, 17/17 Matchstick passing.

## Verification of your three claims

**1. `isAllowedImplementation` matches the canonical staking allow-list — confirmed for all 4 v1 networks.**

Compared `subgraphs/pearl-transactions/src/constants.ts` against `subgraphs/staking/src/utils.ts:176-219`:

| Network | pearl-transactions | staking subgraph | ✓ |
|---|---|---|---|
| gnosis   | `0xEa00be6690a871827fAfD705440D20dd75e67AB1` | same | ✅ |
| matic    | `0x4aba1Cf7a39a51D75cBa789f5f21cf4882162519` | same | ✅ |
| optimism | `0x63C2c53c09dE534Dd3bc0b7771bf976070936bAC` | same | ✅ |
| base     | `0xEB5638eefE289691EcE01943f768EDBF96258a80` | same | ✅ |

Side-note: pearl-transactions uses `implementation.equals(allowed[i])` while the staking subgraph still does `.toHexString() == .toHexString()` — the bytes-level compare here is the more robust choice (no hex-formatting edge cases). Staking subgraph could be tightened in a follow-up but not blocking here.

The Phase-1b scope deliberately excludes `arbitrum-one` / `celo` / `mainnet` — confirmed those aren't in pearl-transactions's network set (4 v1 networks only), so the omission is correct.

**2. `StakingFactory` address + startBlock match `subgraphs/staking/networks.json` — confirmed exact-match on all 4 networks.**

| Network | pearl-transactions addr / startBlock | staking subgraph | ✓ |
|---|---|---|---|
| gnosis   | `0xb022…4700` / `35206806`  | same | ✅ |
| matic    | `0x46C0…b461` / `62213142`  | same | ✅ |
| optimism | `0xa45E…8eD8` / `124618633` | same | ✅ |
| base     | `0x1cEe…f63a` / `17310019`  | same | ✅ |

**3. `_stake` flow ↔ `ServiceStaked.owner = Master Safe` — confirmed (cross-reference from the #131 verification).**

`StakingBase._stake` L865 `safeTransferFrom(msg.sender → address(this))` + L831 `sInfo.owner = msg.sender`. Since the Pearl Master Safe is what calls `stake()`, `ServiceStaked.owner` is the Master Safe, and the round-trip on unstake (L946 `transferFrom(address(this) → msg.sender)`) returns the NFT to it. The link-write in `handleServiceStaked` is on the canonical discovery path.

## Conflict resolution — the dual-guard pattern is the right call

The resolved `handleServiceNftTransfer` (final L122-138 of `src/service-registry.ts`) keeps both:

```ts
if (to.equals(Address.zero())) return;       // mint / burn
if (isStakingContract(to)) return;           // Phase 1b: allowlist short-circuit (no eth_call)
const masterSafe = getOrCreateMasterSafe(to, event);   // defence-in-depth: getOwners() probe
if (masterSafe != null) { ... }              // skip non-Safe recipients
```

This is correct and worth keeping. The allowlist fast-path covers the steady-state Pearl flow (Master Safe ↔ known proxy ↔ burn), and the `getOwners()` revert-probe is the safety net for the two windows where the allowlist might miss:
- **StakingFactory startBlock catch-up gap.** If a service is staked on a proxy whose `InstanceCreated` event hasn't been indexed yet (older block than `StakingFactory.startBlock`), the entity won't exist; the probe stops the phantom Master Safe.
- **Untracked / older-factory proxies.** Any proxy spawned by a factory outside the current allow-list / startBlock window similarly falls through to the probe.

For an indexer with org-wide blast radius, defence-in-depth here is cheap insurance against future factory churn.

## `getOrCreateMasterSafe` null-check at `handleServiceStaked` — needed and correct

Two reasons it's the right fix:
- **Compile-required.** Phase-1a flipped the return type to `MasterSafe | null`, so the prior `masterSafe.id` deref would have been a typecheck error.
- **Semantically correct.** `ServiceStaked.owner` should always be a Master Safe for a Pearl service, but a non-Pearl service staked on a Pearl-relevant staking contract (EOA-owned, foreign-Safe-owned) would fall through and not crash. The guard preserves "best-effort link, never crash" — same policy as the NFT handler.

The new test `"NFT-Transfer to a known StakingContract does NOT call getOwners()"` is elegant: by omitting the `getOwners()` mock, it proves the allow-list short-circuits via matchstick's missing-mock crash signal. ✅

## Nits / non-blocking follow-ups

None worth holding the PR for. One micro-suggestion if you're touching this area later: the `subgraphs/staking` allow-list comparator using `.toHexString()` is a known fragility (we already fixed it on the pearl-transactions side); could be a one-line PR against the staking subgraph at some point.

Merge order remains: #134 → #130 → #131 → **#132** → hold #133 pending the AC #3 decision.

rajat2502:
Thanks for the detailed cross-check @kupermind 🙏 Glad the dual-guard resolution and the null-check both hold up.

Agreed on the nit: the `subgraphs/staking` `isAllowedImplementation` `.toHexString()` comparator is a separate-repo concern — I'll leave it out of this PR and it can be a one-line follow-up against the staking subgraph (tracking it, not blocking here).

Noted on merge order, and I'll **hold #133 pending the AC #3 decision** as you flagged.

rajat2502 (dismissed):

DIvyaNautiyal07 (dismissed):

rajat2502 (approved):

DIvyaNautiyal07 (approved):

PR #133, opened 2026-05-27
phase-2a(pearl-transactions): OLAS Transfer + Safe template + classifyTransfer + AgentFundingEvent + SAFE_SETUP_TRANSFER baseline

kupermind:
## Summary

Step 7 of [`IMPLEMENTATION-PLAN.md`](https://github.com/valory-xyz/autonolas-subgraph-studio/blob/main/subgraphs/pearl-funds/IMPLEMENTATION-PLAN.md) §8. Rebased onto `main` (phase-1a #131 + phase-1b #132 merged).

Brings up the raw token / native ledger: OLAS Transfer data source, `Safe` dynamic template, the `classifyTransfer` engine, `Token` / `TrackedSafe` / `TrackedEOA` / `TokenBalance` / `AgentFundingEvent` entities, and the **`SAFE_SETUP_TRANSFER`** anchor for the wallet UI's "Setup complete" row — implemented per the finalized **AC #3 / Path A** decision.

## What landed

**Schema (additive):** `Token`, `TrackedSafe`, `TrackedEOA`, `TokenBalance`, `AgentFundingEvent`; `MasterSafe.setupTransferSeen` flag + `historyFloorBlock` / `historyFloorTimestamp`; `FundsMovement.agentFundingEvent` / `stakingContract` / `epoch`; new `FundsCategory` variants (`SAFE_SETUP_TRANSFER`, `MASTER_FUNDING_IN`, `MASTER_WITHDRAWAL`, `MASTER_TO_AGENT`, `AGENT_TO_MASTER`, `AGENT_TO_APP`, `APP_TO_AGENT`, `OTHER`).

**Data sources:** OLAS ERC-20 `Transfer` per network; `Safe` dynamic template (`SafeReceived`, `ExecutionSuccess`/`ExecutionFromModuleSuccess` no-op placeholders per §6.2, `AddedOwner`/`RemovedOwner`/`ChangedThreshold`). `ServiceRegistryL2` + `StakingProxy` extended to declare the `ERC20` ABI for `getOrCreateMasterSafe`'s context.

**Handlers:**
- `src/erc20.ts` `handleOlasTransfer` — chain-wide OLAS Transfer, `TrackedAddress` in-handler filter, classify → emit `FundsMovement(source=RAW_TRANSFER)`, group `MASTER_TO_AGENT` under `AgentFundingEvent`, update `TokenBalance`.
- `src/safe.ts` — `SafeReceived` (native inbound), owner-list maintenance (`AddedOwner`/`RemovedOwner`/`ChangedThreshold`).
- `classifyTransfer` (`src/utils.ts`) — routes (from, to) against `TrackedSafe` / `TrackedEOA` / `StakingContract`, most-specific first; returns null for untracked counterparties (row dropped).

## AC #3 / Path A — "Setup complete"

Implemented per the finalized **Path A** decision (plan Rev. 5, §6.2):

- The subgraph emits **no opening-balance row**. At first sighting it records `MasterSafe.historyFloorBlock` / `historyFloorTimestamp`; the frontend renders a "History starts here" divider there and fetches opening balances via archive RPC (`balanceOf` for ERC-20, `eth_getBalance` for native).
- **`SAFE_SETUP_TRANSFER`** is the **first live Master-EOA → Master-Safe inbound hop** observed after first sighting (gated on `setupTransferSeen`); subsequent inbound hops are `MASTER_FUNDING_IN`.
- `TokenBalance` is **delta-only** (seeded from live transfers, not an opening snapshot) — the frontend owns the opening balance, the subgraph owns the movements since.

> Note: the earlier revisions of this branch carried an eth_call `OPENING_BALANCE` baseline (a `balanceOf` snapshot per tracked ERC-20 + a native marker). That contradicted Path A and has been **removed** — see the final commit. The `OPENING_BALANCE` enum value, its constant, the emit functions, and the related test expectations are gone.

**Known accepted gap — native → Agent EOA:** native transfers to a plain EOA emit no log; the Agent-EOA gas-funding leg is permanently unobservable from event-based indexing on these chains. Accepted for v1, documented in §6.2 / the subgraph CLAUDE.md.

## Tests

`tests/phase-2a.test.ts` covers classify routing (setup-transfer, funding-in, master→agent grouping, agent→master sweep, withdrawal, untracked-drop, OTHER, app in/out), native `SafeReceived`, owner-list maintenance, `TokenBalance` accumulation, and wrapped-native routing. **36/36 Matchstick passing across all phases** (10 phase-1a + 9 phase-1b + 17 phase-2a); codegen clean.

## Test plan

- [x] `yarn install` + `generate-manifests` + `codegen` + `test` — 36/36 passing.
- [x] `node scripts/audit-install-hooks.mjs` — OK.
- [x] Root `Dependency audit` green on the rebased branch.

## Follow-up

| Step | Adds |
|---|---|
| 5 — Verify Phase 1+2a on Studio | spot-check end-to-end against a known Pearl service |
| 6 — graph-node `Template.createWithContext` startBlock verification | Open Q #6 (§6.2 option 1 — not needed for Path A) |
| 8 — Phase 2b | USDC / USDC.e Transfer data sources (benchmark + product-gated, §6.3) |


rajat2502:
Flagging before this gets reviewed/rebased: **the `SAFE_SETUP_TRANSFER` implementation here conflicts with the finalized AC #3 / Path A decision (plan Rev. 5, §6.2).** This is a rework, not just a rebase — and it's why #132's review thread tags #133 as "hold pending the AC #3 decision."

### What Rev. 5 / Path A says
- "Setup complete" opening balances are rendered **frontend-side** via archive RPC (`balanceOf` / `eth_getBalance` at `MasterSafe.historyFloorBlock`). **"No subgraph change needed."** The `OPENING_BALANCE` snapshot was explicitly removed — the plan states it "does not satisfy AC #3 as written."
- `SAFE_SETUP_TRANSFER` is redefined as the **first _live_ Master-EOA → Master-Safe inbound ERC-20/native hop** observed after first sighting (detected via the transient `setupTransferSeen` flag in the transfer handler). Subsequent inbound hops → `MASTER_FUNDING_IN`.

### What this PR currently does (per the description)
> `getOrCreateMasterSafe` queries `OLAS.balanceOf(masterSafe)` and emits a synthetic baseline row … marks `setupTransferSeen = true` so subsequent live hops classify as `MASTER_FUNDING_IN`.

Two issues against Path A:
1. It emits a **balance-snapshot row at first sighting** — that's the `OPENING_BALANCE` concept Rev. 5 deleted and reassigned to the frontend. The subgraph shouldn't emit it.
2. It sets `setupTransferSeen = true` **at first sighting**, so the first real live inbound hop is labelled `MASTER_FUNDING_IN` — meaning a true `SAFE_SETUP_TRANSFER` live-transfer row is **never** emitted.

Net: this resurrects the eth_call-baseline approach that Path A rejected.

### Suggested rework (once #131 + #132 land and this rebases onto `main`)
- `getOrCreateMasterSafe`: drop the `OLAS.balanceOf` call + synthetic baseline row; set `setupTransferSeen = false`; record `historyFloorBlock` (new field, distinct from `firstSeenBlock`).
- OLAS-Transfer + Safe-native handlers: detect the first live Master-EOA → Master-Safe inbound hop, flip the flag, emit **that hop** as `SAFE_SETUP_TRANSFER`; subsequent → `MASTER_FUNDING_IN`.
- Expose `historyFloorBlock` for the frontend's Path A balance reads.
- Tests: convert the SAFE_SETUP_TRANSFER cases from "baseline snapshot" to "first live inbound hop."

Recommend holding the rebase until #131/#132 merge (base is moving), then doing the Path A rework as one focused change rather than reworking against a shifting base.

kupermind:
Reviewed against the current plan + checked the code mechanics. The code change itself is clean, but **it presupposes a "Rev. 5 / Path A" decision that hasn't actually landed yet** — flagging the process gap before this merges.

## What changed mechanically (✅ correct as drafted)

Rajat's commit cleanly reverses the Rev. 4 `OPENING_BALANCE` mechanism:

- `schema.graphql` — drops `OPENING_BALANCE` from the `FundsCategory` enum; rewords `SAFE_SETUP_TRANSFER` to "first LIVE Master EOA → Master Safe hop after first sighting"; keeps `MasterSafe.historyFloorBlock` + `historyFloorTimestamp` as the FE archive-RPC anchor.
- `src/utils.ts` — removes `emitOpeningBalanceRows` + `emitOpeningBalanceForToken` (~110 LoC, plus the `ERC20` import); `getOrCreateMasterSafe` no longer calls `OLAS.balanceOf` / `WrappedNative.balanceOf` and no longer flips `setupTransferSeen`. `markSetupTransferSeen` comment cleaned to drop the eth-call-baseline reference.
- `src/constants.ts` — drops `CATEGORY_OPENING_BALANCE`.
- Tests — drops the `mockOlasBalanceOf` / `mockWrappedNativeBalanceOf` helpers from both 1a + 1b suites (no longer needed); rewrites the relevant entity-count assertions (e.g. `assert.entityCount("FundsMovement", 4)` → `1`, `8` → `5`). The Phase-2a "OPENING_BALANCE does NOT pre-empt SAFE_SETUP_TRANSFER" test renamed to "first live Master EOA → Master Safe inbound is SAFE_SETUP_TRANSFER, subsequent are MASTER_FUNDING_IN" — same semantics, just no longer requires the Rev. 4 baseline.

Net: 42 insertions / 241 deletions. CI all-green ✅. The mechanical correctness is solid — `setupTransferSeen` flag, classifier, AgentFundingEvent grouping are all preserved; only the synthetic baseline path is gone.

## The blocker — there is no Rev. 5

The commit message references **"plan Rev. 5, §6.2"** and the comments quote what "Rev. 5 / Path A says". But:

1. **The latest plan commit on `main` is Rev. 4** (`391a2f1 docs(pearl-funds): Rev. 4 — §11 #6 RESOLVED (NOT SUPPORTED), §6.2 Path B is dead`). No Rev. 5 commit exists anywhere in the repo.
2. **Rev. 4 explicitly records Path A as a *recommendation*, not a decision.** Direct quote from `subgraphs/pearl-funds/IMPLEMENTATION-PLAN.md` §6.2 line 950:

   > "Path A is the realistic answer. Path B is dead (verified upstream); Path C options are all heavier re-architectures than Rev. 4. **Needs Pearl product + FE sign-off** that the AC #3 wording can be relaxed to 'Setup complete = `historyFloorBlock` + the `OPENING_BALANCE` row set, rendered FE-side'. If product holds AC #3 as written, Path C-1 (discovery + graft + static dataSources) is the lowest-pain on-chain workaround."

3. The Path A row in the §6.2 decision table currently reads **"None — Rev. 4 is the contract"** (line 908). Until a `docs(pearl-funds): Rev. 5` commit flips that, the plan-of-record is Rev. 4 with `OPENING_BALANCE`.

So this commit makes the code march to Path A while the docs still record Path A as conditional on PM/FE sign-off. **The hold tag on this PR is doing real work** — it's gated on the PM/FE conversation that the original verdict comment ([PR #130 → 4556589761](https://github.com/valory-xyz/autonolas-subgraph-studio/pull/130#issuecomment-4556589761)) called for.

## What needs to happen before this merges

Pick one:

- **(a) Confirm Path A with PM/FE first**, then land a `docs(pearl-funds): Rev. 5` plan commit on `docs/pearl-funds-plan` that:
  - flips §6.2 Path A row to "**Decided** — Rev. 5 is the contract"
  - deletes the `OPENING_BALANCE` references throughout (including §6.5 schema notes, §6.6 "What Phase 2 answers", §10 test scope, §11 open Qs)
  - updates the "Last updated" line
  
  …then this PR merges cleanly. Order: PM/FE call → Rev. 5 plan PR merges → #133 merges.

- **(b) If product holds AC #3 as written**, this PR is the wrong shape entirely — Path C-1 (discovery + graft + static dataSources) is the called-for rework, not the OPENING_BALANCE delete. That's substantially more work and the current commit would need to be backed out.

In either case: **don't merge #133 ahead of the PM/FE decision.** Holding makes the work cheap to revert if (b) wins; merging now bakes in (a) by default.

## Smaller observations (non-blocking even if Path A is confirmed)

- **`MasterSafe.historyFloorBlock` is now load-bearing for the FE.** It's the only consumer-visible anchor under Path A. Worth a sentence in the schema comment that this is the block the FE uses for the `balanceOf` / `eth_getBalance` archive-RPC reads — it's already there ("Anchor fields the wallet UI uses…") but emphasising "single source of truth for opening-balance reads" makes the contract obvious to a future maintainer.
- **The native pre-discovery gap remains real and now has no UI hook.** Rev. 4's zero-amount native marker row at least gave the FE a row to attach a "native pre-discovery balance unknown" caption to; with Path A the FE has to surface that caption from `historyFloorBlock` metadata alone. Worth a Rev. 5 plan paragraph confirming the FE team is OK with this.
- **No CLAUDE.md update yet** — fine for now (deferred to Phase 9 docs PR per the original sequencing), but the file currently describes the Phase 1a state, not even Rev. 4 Phase 2a, so it'll need a sweep.
- **No `IMPLEMENTATION-PLAN.md` update in this commit.** Path A is a *plan-and-code change*, not just a code change; the Rev. 5 commit should land first (or in the same PR series) so the plan and code don't diverge.

## TL;DR

Code is clean and correctly implements Path A. The decision Path A presupposes hasn't been confirmed (no Rev. 5, plan still says "needs PM/FE sign-off"). Hold this PR until the PM/FE conversation happens and Rev. 5 lands. If Path A wins, this merges as-is; if AC #3 is held literally, the rework is C-1 not OPENING_BALANCE-delete.

Merge order, updated: #134 → #130 → #131 → #132 → **(PM/FE call → `docs(pearl-funds): Rev. 5` PR) → #133**.

rajat2502:
Thanks @kupermind — you're right that the decision needs to be the plan-of-record, not just presupposed by the code. Confirming the process gap and closing it:

**PM/FE sign-off happened.** Product + FE agreed to **fetch opening balances frontend-side** (archive-RPC `balanceOf` / `eth_getBalance` at `historyFloorBlock`) — i.e. Path A is the decision, AC #3's "Setup complete" is relaxed to "`historyFloorBlock` anchor + FE-rendered opening balances." So this is case **(a)**, not C-1; the OPENING_BALANCE-delete is the right shape and the code stands as-is.

**Closing the doc gap:** the Rev. 5 plan text already exists on **#129** (`docs/pearl-funds-plan`) — *"AC #3 confirmed as Path A, product decisions finalised"* — it just hasn't merged yet (the plan isn't on `main` at all). Plan of action, matching your merge order:
> … → #132 → **#129 (Rev. 5 plan) merges** → #133

I'll make sure the Rev. 5 plan PR flips the §6.2 Path A row to **Decided** and strips the residual `OPENING_BALANCE` references (§6.5 / §6.6 / §10 / §11) before it lands, so plan and code don't diverge.

**Non-blocking nits addressed (1473acd):** bumped the `historyFloorBlock` schema comment to state explicitly it's the *single source of truth* for the FE's opening-balance reads, and that the native pre-discovery "unknown before this block" caption hangs off this anchor (your two schema/UI-hook observations). CLAUDE.md catch-up stays deferred to the Phase 9 docs sweep as you noted.

Net: holding #133 behind #129 as you suggest; no code rework needed.

kupermind:
Followed up on the addressed comments. All clear from my side — verified against the actual artifacts.

## ✅ Process gap closed — Rev. 5 plan PR exists and says what's claimed

Pulled `docs/pearl-funds-plan` and diff'd against `main` (Rev. 4). Two Rev. 5 commits on **PR #129** (`2e698a0b` + `70f51050`); they substantively close every concern I raised:

- **§6.2 Path A formally decided.** The old decision table is gone, replaced with: *"AC #3 — resolved (Rev. 5, 2026-05-29): Path A confirmed by product. 'Setup complete' is rendered frontend-side… No subgraph change needed."* The §6.2 "Path C workarounds" section is also stripped (correct — they no longer apply).
- **`OPENING_BALANCE` excised from §6.5 schema** + reworded in §5/§6 commentary; "Pre-template-spawn transfers" rewritten with "Rev. 5 decision: Opening balances are delegated to the frontend" and the explicit archive-RPC contract on `historyFloorBlock`.
- **§11 Open Q #5 (subgraph name)** struck through and resolved (`pearl-transactions`).
- **§11 Open Q #6 (graph-node backdated startBlock)** consequence rewritten to point at the Path A resolution rather than the prior recommendation.
- **§11 Open Q #8 (pre-Master-Safe Master EOA history)** resolved as option (a) — `historyFloorBlock` is the contract; native → Agent EOA confirmed as accepted v1 gap.
- **"Last updated" header** flipped to `2026-05-29 (Rev. 5 — product decisions finalised…)`.

So Rajat's earlier claim that "Rev. 5 already exists on #129" checks out exactly. Plan-of-record and code will be consistent the moment #129 merges.

## ✅ `1473acd` nit commit looks good

Comment-only schema bump on `MasterSafe.historyFloorBlock`. Now explicitly labels it the **SINGLE SOURCE OF TRUTH** for opening balances under Path A, calls out the archive-RPC contract (`balanceOf` for ERC-20, `eth_getBalance` for native), and notes the native pre-discovery "unknown before this block" caption hangs off this anchor — i.e. both observations from the prior review. No logic change. ✅

## ⚠️ One CI note (not blocking — not yours)

The latest CI run (`26757439912`) shows `test / liquidity-l2 FAILURE` and cascading `All checks passed FAILURE`. **`test / pearl-transactions` is green**, lockfile validation for pearl-transactions is green — the failure is an **upstream `graph-cli` flake unrelated to this PR**:

```
Fetching latest version tag...
Error: Failed to get matchstick binary: Invalid version. Must be a string.
   Got type "undefined".
```

This is the matchstick-binary downloader in `@graphprotocol/graph-cli` choking on an upstream version-tag response — same hash a rerun should clear, and it only hit `liquidity-l2` (transient network/registry blip). The previous run on `f4a5b5c3` (right before the nit commit) was all-green, and `1473acd` is comment-only, so there is zero chance this is caused by your changes. A "Re-run failed jobs" should turn the rollup green; if it doesn't, the failure belongs to a separate `liquidity-l2` ticket, not #133.

## Updated merge order (unchanged from yesterday)

> #134 → #130 → #131 → #132 → **#129 (Rev. 5 plan) → #133**

#129 needs to land before #133 so plan and code don't diverge. Once #129 is in, this PR is good to merge as-is — no further code changes required.

Thanks for the quick turnaround.

DIvyaNautiyal07 (approved):
Approved. The classifier ordering, the AC #3 / Path A pivot, and the Safe template wiring all look right. The historyFloorBlock comment clarification in the latest commit makes the consumer contract a lot easier to follow.

A few nice-to-haves for this PR or a follow-up:

- `src/utils.ts:230-248` — in `getOrCreateAgentSafe`, the `upsertTrackedSafe` call is guarded on `service.masterSafe !== null`, but `SafeTemplate.create(address)` and the operators loop run unconditionally. If `service.masterSafe` is null at this point (partial historical sync, replay edge case), the Safe gets a template listener but no `TrackedSafe` entity, so any later OLAS Transfer through this agent falls through `classifyTransfer` to null and is silently dropped. Either gate the whole block on the same condition or log a warning.

- `src/utils.ts:889, 905` + `schema.graphql:272` — `AgentFundingEvent.totalOlasAmount` is set from every non-native ERC-20, not just OLAS. The PR description says Phase 2b adds USDC / USDC.e Transfer data sources next, at which point the same handler will sum 6-decimal USDC into this 18-decimal OLAS field. Rename to `totalErc20Amount` (single-token-only) or add a per-token breakdown before Phase 2b lands.

- `src/safe.ts:80-91 vs 99-116` — the block comment for `handleSafeExecutionSuccess` says "We record a zero-amount placeholder row so the consumer wallet UI can show 'Safe executed an outbound tx'". `emitNativeOutPlaceholder` then loads `MasterSafe.load(safeAddr)`, discards the result, and returns without emitting anything. Line 117 says "intentionally a no-op for v1." Either drop the misleading promise from the block comment and remove the dead load, or actually emit the placeholder row described.

- `subgraphs/pearl-transactions/CLAUDE.md` — the file wasn't updated for Phase 2a. Header still says "Currently a scaffold only", "Current state — Phase 1a" is the only state section, and the roadmap row for step 7 still describes Phase 2a as upcoming. A reader following the docs won't know Phase 2a is live.

- `src/utils.ts:783` — `SAFE_SETUP_TRANSFER` only fires when `fromEOA != null && fromEOA.role == "MASTER_EOA"`. If `getOwners()` returns a non-Pearl-shaped ordering and the real Master EOA isn't `owners[0]`, no `TrackedEOA` is seeded for the real Master EOA. Every inbound from it classifies as `MASTER_FUNDING_IN`, the flag is never flipped, and the "History starts here" anchor row is never produced. A `log.warning` on this miss would at least make it observable.

- `src/utils.ts:832-845` — the OTHER fallback resolves `masterRef` only from `fromEOA.masterSafe` / `toEOA.masterSafe`. A transfer between two Agent Safes (both `AGENT` role, both tracked, different services) lands in OTHER with `masterRef = null` and is orphaned. Extend resolution to fall back to `fromMaster.masterSafe` / `toMaster.masterSafe`.

- `src/utils.ts:854-858` — `markSetupTransferSeen` silently returns on the `ms == null` path. That branch indicates the caller wrote a SAFE_SETUP_TRANSFER row against a MasterSafe ID that doesn't exist — worth a `log.warning` since the consequence (flag never flips, next identical transfer also classifies as SAFE_SETUP_TRANSFER) is silently wrong.

- `schema.graphql:230, 240` — `TrackedSafe.role` and `TrackedEOA.role` are freeform `String!` with valid values only in comments, and `classifyTransfer` does string equality against the same literals at five-plus sites. A typo at any callsite gives silently wrong classification. Same pattern flagged on `Service.state` in PR #131 — would be cleaner as `enum TrackedSafeRole { MASTER AGENT }` and `enum TrackedEoaRole { MASTER_EOA AGENT_EOA }`.

- `tests/phase-2a.test.ts:317-345` — the MASTER_TO_AGENT test only asserts `entityCount("AgentFundingEvent", 1)` and the FundsMovement categories. It never asserts `totalOlasAmount` or `totalNativeAmount`. A mutation that removed `addToAgentFundingEvent` or swapped the `isNative` flag would pass every test.

- `tests/phase-2a.test.ts:384-402` — the only `SafeReceived` test seeds with `setupTransferSeen=true`, so the SAFE_SETUP_TRANSFER + flag-flip path inside `handleSafeReceived` is never exercised. Add a test that fires SafeReceived from `MASTER_EOA` with the flag false and asserts the native-side setup-transfer behavior.

- `tests/phase-2a.test.ts:404-439` — the AddedOwner / RemovedOwner test only adds `NEW_OWNER` (not already in owners) and removes `BACKUP_EOA` (not `owners[0]`). The dedupe guard in `handleSafeAddedOwner` and the `masterEoa` promotion branch in `handleSafeRemovedOwner` are both untested.

- `src/erc20.ts:87-117` — the TokenBalance update is split into three `if` blocks where `MASTER_TO_AGENT` appears in two of them and the trailing comment on line 117 ("AGENT_TO_MASTER also updates Master side; handled above for 'to'") points "above" to a block that's actually the first one. Cleaner as one `debit-from` block plus one `credit-to` block where each category appears once on whichever side(s) it needs.

- `src/utils.ts:645` — `bal.balance = isDelta ? amount : amount` — both branches are `amount`. Happens to be correct on a new entity (delta-from-0 equals absolute-set), but the ternary is redundant. Just write `bal.balance = amount;`.

- `src/safe.ts:108` — `const masterSafe = MasterSafe.load(safeAddr);` inside `emitNativeOutPlaceholder` is dead — the variable is never read. Either remove the load or use it for the documented placeholder behavior.

- `src/utils.ts:129`, `src/constants.ts:91`, `tests/phase-2a.test.ts:596` — "Rev. 4" / "Rev. 5" internal revision tags mean nothing to a future reader who hasn't seen the in-flight PR history. Drop the tags; keep the attribution to the @Tanya-atatakai review where useful.

- `src/erc20.ts:29-85` + `src/safe.ts:28-78` — both handlers run the same ~15-line classify-then-populate-FundsMovement sequence, differing only by `row.token`. A `populateFundsMovement(row, classification, event, token | null)` helper would unify them.

rajat2502 (approved):

PR #138, opened 2026-06-01
phase-2b(pearl-transactions): stablecoin Transfer data sources (USDC / USDC.e / pUSD)

rajat2502:
## Summary

Phase 2b (plan §8 step 8) — adds the **stablecoin `Transfer` data sources** the Pearl wallet surfaces, completing the raw-token ledger. Reuses the Phase-2a generic `handleErc20Transfer` + `classifyTransfer` unchanged (the row's `token` comes from `event.address`), so this is almost entirely manifest/config + token-metadata.

Stacked on Phase-2a (#133, merged) → based on `main`.

## Token set (sourced from the Operate app `frontend/config/tokens.ts`)

All **6-decimal**:

| Chain | Tokens |
|---|---|
| gnosis | USDC `0xDDAfbb50…`, USDC.e `0x2a22f9c3…` |
| matic | USDC `0x3c499c54…`, USDC.e `0x2791Bca1…`, pUSD `0xC011a7E1…` |
| optimism | USDC `0x0b2C639c…`, USDC.e `0x7F5c764c…` |
| base | USDC `0x833589fC…` |

> **Note:** plan §6.3 assumed Polystrat funds in USDC.e on Polygon, but the current app config shows it uses **USDC + pUSD** — the token set follows the app as source of truth, not the (stale on this point) plan.

## What changed

- **`scripts/generate-manifests.js`** — optional per-network `erc20Tokens` array expands into N `Transfer` data sources via a new `{{ erc20TokenDataSources }}` template marker. **Guarded + additive**: no-op for templates without the marker and networks without the array. Verified the other 12 subgraphs generate **byte-identical** output (regenerated staking / tokenomics-l2 / liquidity-l2 / service-registry — the only diff, service-registry, reproduces with the *original* script too = pre-existing drift, not this change).
- **`networks.json`** — `erc20Tokens` arrays per network. Each data source's `startBlock` = that chain's `ServiceRegistryL2` block (provably-safe lower bound — no Pearl Safe predates it, per plan Open Q #7) — no per-token deploy-block lookup needed.
- **`constants.ts`** — `getStablecoinSymbol(network, address)` resolver; **`utils.ts`** `getOrCreateToken` tags USDC / USDC.e / pUSD at **6 decimals** (others stay 18).
- **`tests/phase-2a.test.ts`** — Gnosis USDC.e @ 6 decimals + Polygon pUSD @ 6 decimals (network switch). The generic classify routing is already covered by the Phase-2a suite.

## Test plan

- [x] `yarn install` + `generate-manifests` + `codegen` + `test` — **38/38 Matchstick passing**.
- [x] Verified the shared `generate-manifests.js` change is a no-op for the other 12 subgraphs.
- [x] `gitleaks` / lockfile-lint — no issues.

## Deferred (not in this PR)

- **§6.3a Polygon sync benchmark** — wanted a throwaway USDC.e sync measurement before committing on-chain; deferred because Studio deploy isn't provisioned yet. On-chain path chosen per product. **Revisit if Polygon full-sync proves too slow** (USDC.e on Polygon is the §2.2 cost hotspot).
- **CLAUDE.md / README full refresh** — Phase 9 docs sweep (only the Phase 2b table row is updated here).

Related: plan #129, scaffold #130, phase-1a #131, phase-1b #132, phase-2a #133.

### DIFF
--- a/subgraphs/pearl-transactions/src/constants.ts
+++ b/subgraphs/pearl-transactions/src/constants.ts
@@ -136,7 +136,9 @@ export function getStablecoinSymbol(
   network: string,
   address: Address
 ): string | null {
-  const a = address.toHexString();
+  // Lowercase explicitly: the address literals below are lowercase, and
+  // this drops a hidden dependency on graph-ts toHexString() casing.
+  const a = address.toHexString().toLowerCase();
   if (network == "gnosis" || network == "xdai") {
     if (a == "0xddafbb505ad214d7b80b1f830fccc89b60fb7a83") return "USDC";
     if (a == "0x2a22f9c3b484c3629090feed35f17ff8f88f76f0") return "USDC.e";
--- a/subgraphs/pearl-transactions/src/constants.ts
+++ b/subgraphs/pearl-transactions/src/constants.ts
@@ -126,6 +126,33 @@ export function getWrappedNativeSymbol(network: string): string {
   return "WNATIVE";
 }
 
+// getStablecoinSymbol — Phase 2b stablecoin metadata resolver. Returns
+// the display symbol for a known per-network stablecoin (all 6 decimals),
+// or null if `address` isn't a tracked stablecoin. Addresses mirror the
+// `erc20Tokens` set in networks.json (sourced from the Operate app's
+// frontend/config/tokens.ts). All comparisons are lowercase-hex
+// (graph-ts toHexString() is lowercase).
+export function getStablecoinSymbol(
+  network: string,
+  address: Address
+): string | null {
+  const a = address.toHexString();
+  if (network == "gnosis" || network == "xdai") {
+    if (a == "0xddafbb505ad214d7b80b1f830fccc89b60fb7a83") return "USDC";
+    if (a == "0x2a22f9c3b484c3629090feed35f17ff8f88f76f0") return "USDC.e";
+  } else if (network == "matic" || network == "polygon") {
+    if (a == "0x3c499c542cef5e3811e1192ce70d8cc03d5c3359") return "USDC";
+    if (a == "0x2791bca1f2de4661ed88a30c99a7a9449aa84174") return "USDC.e";
+    if (a == "0xc011a7e12a19f7b1f670d46f03b03f3342e82dfb") return "pUSD";
+  } else if (network == "optimism") {
+    if (a == "0x0b2c639c533813f4aa9d7837caf62653d097ff85") return "USDC";
+    if (a == "0x7f5c764cbc14f9669b88837ca1490cca17c31607") return "USDC.e";
+  } else if (network == "base") {
+    if (a == "0x833589fcd6edb6e08f4c7c32d4f71b54bda02913") return "USDC";
+  }
+  return null;
+}
+
 // isAllowedImplementation — the Olas staking ecosystem allows multiple
 // StakingProxy implementations but pearl-transactions only indexes
 // proxies whose implementation appears on this per-network allow-list.
--- a/subgraphs/pearl-transactions/src/utils.ts
+++ b/subgraphs/pearl-transactions/src/utils.ts
@@ -37,6 +37,7 @@ import {
   SOURCE_SEMANTIC,
   getOlasAddress,
   getSrtuAddress,
+  getStablecoinSymbol,
   getWrappedNativeAddress,
   getWrappedNativeSymbol,
 } from "./constants";
@@ -617,10 +618,17 @@ export function getOrCreateToken(tokenAddress: Address): Token {
     token.symbol = getWrappedNativeSymbol(network);
     token.decimals = 18;
   } else {
-    // Fallback for any unknown token that shows up via classifyTransfer
-    // (shouldn't happen given our data-source set, but defensive).
-    token.symbol = "UNKNOWN";
-    token.decimals = 18;
+    const stablecoin = getStablecoinSymbol(network, tokenAddress);
+    if (stablecoin !== null) {
+      // Phase 2b stablecoins (USDC / USDC.e / pUSD) — all 6 decimals.
+      token.symbol = stablecoin;
+      token.decimals = 6;
+    } else {
+      // Fallback for any unknown token that shows up via classifyTransfer
+      // (shouldn't happen given our data-source set, but defensive).
+      token.symbol = "UNKNOWN";
+      token.decimals = 18;
+    }
   }
   token.save();
   return token;
--- a/subgraphs/pearl-transactions/src/utils.ts
+++ b/subgraphs/pearl-transactions/src/utils.ts
@@ -600,12 +600,17 @@ export function upsertTrackedEOA(
   tracked.save();
 }
 
-// getOrCreateToken — Phase 2a tokens are OLAS (always 18 decimals,
-// symbol "OLAS") and the per-chain wrapped native (always 18 decimals;
-// symbol per getWrappedNativeSymbol). Phase 2b will extend with USDC
-// (6 decimals) and USDC.e (6 decimals). Token metadata is hardcoded
-// rather than queried because the ERC20Detailed ABI in this repo
-// doesn't include symbol() and Pearl's token set is small + known.
+// getOrCreateToken — resolves token metadata for the indexed token set.
+// Metadata is hardcoded (not queried) because the ERC20Detailed ABI in
+// this repo doesn't include symbol()/decimals() and Pearl's token set is
+// small + known:
+//   - OLAS — 18 decimals (per-chain address via getOlasAddress).
+//   - wrapped native — 18 decimals (symbol per getWrappedNativeSymbol).
+//   - stablecoins — USDC / USDC.e / pUSD, all 6 decimals, resolved per
+//     chain via getStablecoinSymbol (the set varies per chain — see
+//     networks.json `erc20Tokens`).
+// First write wins (early return on an existing Token), so a wrong
+// decimals value would persist forever — hence the log.critical guard.
 export function getOrCreateToken(tokenAddress: Address): Token {
   let token = Token.load(tokenAddress);
   if (token != null) return token;
@@ -620,12 +625,21 @@ export function getOrCreateToken(tokenAddress: Address): Token {
   } else {
     const stablecoin = getStablecoinSymbol(network, tokenAddress);
     if (stablecoin !== null) {
-      // Phase 2b stablecoins (USDC / USDC.e / pUSD) — all 6 decimals.
+      // Per-chain stablecoin (USDC / USDC.e / pUSD) — all 6 decimals.
       token.symbol = stablecoin;
       token.decimals = 6;
     } else {
-      // Fallback for any unknown token that shows up via classifyTransfer
-      // (shouldn't happen given our data-source set, but defensive).
+      // Reached here only via a tracked Transfer, so this is an indexed
+      // token with no metadata resolver — almost certainly an
+      // `erc20Tokens` entry in networks.json without a matching
+      // getStablecoinSymbol branch. UNKNOWN/18 would silently misformat a
+      // 6-decimal stablecoin by 10^12 in every consumer, and first-write-
+      // wins makes it permanent. Loud on purpose so it surfaces in
+      // indexing logs, not in the wallet UI.
+      log.critical(
+        "getOrCreateToken: no symbol/decimals resolver for indexed token {} on {} — add a getStablecoinSymbol branch (networks.json `erc20Tokens` and the resolver are out of sync). Falling back to UNKNOWN/18.",
+        [tokenAddress.toHexString(), network]
+      );
       token.symbol = "UNKNOWN";
       token.decimals = 18;
     }
--- a/subgraphs/pearl-transactions/tests/phase-2a.test.ts
+++ b/subgraphs/pearl-transactions/tests/phase-2a.test.ts
@@ -131,6 +131,48 @@ function newErc20TransferAt(
   return e;
 }
 
+// assertStablecoinResolves — seed a Master Safe (setupTransferSeen=true so
+// the inbound classifies as MASTER_FUNDING_IN), fire a Master EOA → Master
+// Safe stablecoin Transfer on `network`, and assert the Token resolves to
+// the expected 6-decimal symbol. File-scoped (not nested in a describe) to
+// avoid the AssemblyScript closure-capture WASM-validator error.
+function assertStablecoinResolves(
+  network: string,
+  tokenStr: string,
+  expectedSymbol: string,
+  salt: i32
+): void {
+  dataSourceMock.setNetwork(network);
+  seedMasterSafe(true);
+  const token = Address.fromString(tokenStr);
+  const tx = mockTx(salt);
+  handleErc20Transfer(
+    newErc20TransferAt(
+      token,
+      MASTER_EOA,
+      MASTER_SAFE,
+      BigInt.fromString("1000000"),
+      tx,
+      0
+    )
+  );
+  const id = tx.concatI32(0);
+  assert.fieldEquals(
+    "FundsMovement",
+    id.toHexString(),
+    "category",
+    "MASTER_FUNDING_IN"
+  );
+  assert.fieldEquals(
+    "FundsMovement",
+    id.toHexString(),
+    "token",
+    token.toHexString()
+  );
+  assert.fieldEquals("Token", token.toHexString(), "symbol", expectedSymbol);
+  assert.fieldEquals("Token", token.toHexString(), "decimals", "6");
+}
+
 function newSafeReceived(
   sender: Address,
   value: BigInt,
@@ -711,51 +753,95 @@ describe("pearl-transactions / Phase 2a — raw OLAS + Safe template", () => {
     );
   });
 
-  // ---- Phase 2b — stablecoins (USDC / USDC.e / pUSD, all 6 decimals) ----
+});
 
-  test("Gnosis USDC.e Transfer routes via the generic handler; Token = USDC.e @ 6 decimals", () => {
-    seedMasterSafe(true);
-    const USDCE_GNOSIS = Address.fromString(
-      "0x2a22f9c3b484c3629090FeED35F17Ff8F88f76F0"
-    );
-    const amount = BigInt.fromString("1000000"); // 1 USDC.e (6 decimals)
-    const tx = mockTx(31);
-    handleErc20Transfer(
-      newErc20TransferAt(USDCE_GNOSIS, MASTER_EOA, MASTER_SAFE, amount, tx, 0)
-    );
-
-    const id = tx.concatI32(0);
-    assert.fieldEquals("FundsMovement", id.toHexString(), "category", "MASTER_FUNDING_IN");
-    assert.fieldEquals(
-      "FundsMovement",
-      id.toHexString(),
-      "token",
-      USDCE_GNOSIS.toHexString()
-    );
-    assert.fieldEquals("Token", USDCE_GNOSIS.toHexString(), "symbol", "USDC.e");
-    assert.fieldEquals("Token", USDCE_GNOSIS.toHexString(), "decimals", "6");
+// ---- Phase 2b — stablecoins (USDC / USDC.e / pUSD; all 6 decimals) ----
+//
+// Covers every (chain, token) tuple in getStablecoinSymbol so a typo in
+// any address literal (or a missing resolver branch) fails loudly here
+// instead of silently mapping to UNKNOWN/18 in production. Each case sets
+// its own network so there's no cross-test ordering dependency.
+describe("pearl-transactions / Phase 2b — stablecoins", () => {
+  beforeEach(() => {
+    clearStore();
+    dataSourceMock.setNetwork("gnosis");
   });
 
-  test("Polygon pUSD Transfer resolves to pUSD @ 6 decimals", () => {
-    dataSourceMock.setNetwork("matic");
-    seedMasterSafe(true);
-    const PUSD_POLYGON = Address.fromString(
-      "0xC011a7E12a19f7B1f670d46F03B03f3342E82DFB"
-    );
-    const amount = BigInt.fromString("2500000"); // 2.5 pUSD (6 decimals)
-    const tx = mockTx(32);
-    handleErc20Transfer(
-      newErc20TransferAt(PUSD_POLYGON, MASTER_EOA, MASTER_SAFE, amount, tx, 0)
-    );
+  afterEach(() => {
+    clearStore();
+    // Reset so a stray network doesn't leak into any later suite.
+    dataSourceMock.setNetwork("gnosis");
+  });
 
-    const id = tx.concatI32(0);
-    assert.fieldEquals(
-      "FundsMovement",
-      id.toHexString(),
-      "token",
-      PUSD_POLYGON.toHexString()
+  test("gnosis USDC resolves @ 6 decimals", () => {
+    assertStablecoinResolves(
+      "gnosis",
+      "0xDDAfbb505ad214D7b80b1f830fcCc89B60fb7A83",
+      "USDC",
+      31
     );
-    assert.fieldEquals("Token", PUSD_POLYGON.toHexString(), "symbol", "pUSD");
-    assert.fieldEquals("Token", PUSD_POLYGON.toHexString(), "decimals", "6");
+  });
+
+  test("gnosis USDC.e resolves @ 6 decimals", () => {
+    assertStablecoinResolves(
+      "gnosis",
+      "0x2a22f9c3b484c3629090FeED35F17Ff8F88f76F0",
+      "USDC.e",
+      32
+    );
+  });
+
+  test("matic USDC resolves @ 6 decimals", () => {
+    assertStablecoinResolves(
+      "matic",
+      "0x3c499c542cEF5E3811e1192ce70d8cC03d5c3359",
+      "USDC",
+      33
+    );
+  });
+
+  test("matic USDC.e resolves @ 6 decimals", () => {
+    assertStablecoinResolves(
+      "matic",
+      "0x2791Bca1f2de4661ED88A30C99A7a9449Aa84174",
+      "USDC.e",
+      34
+    );
+  });
+
+  test("matic pUSD resolves @ 6 decimals", () => {
+    assertStablecoinResolves(
+      "matic",
+      "0xC011a7E12a19f7B1f670d46F03B03f3342E82DFB",
+      "pUSD",
+      35
+    );
+  });
+
+  test("optimism USDC resolves @ 6 decimals", () => {
+    assertStablecoinResolves(
+      "optimism",
+      "0x0b2C639c533813f4Aa9D7837CAf62653d097Ff85",
+      "USDC",
+      36
+    );
+  });
+
+  test("optimism USDC.e resolves @ 6 decimals", () => {
+    assertStablecoinResolves(
+      "optimism",
+      "0x7F5c764cBc14f9669B88837ca1490cCa17c31607",
+      "USDC.e",
+      37
+    );
+  });
+
+  test("base USDC resolves @ 6 decimals", () => {
+    assertStablecoinResolves(
+      "base",
+      "0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913",
+      "USDC",
+      38
+    );
   });
 });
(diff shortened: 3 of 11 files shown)


subgraphs/pearl-transactions/src/utils.ts
```diff
@@ -617,10 +618,17 @@ export function getOrCreateToken(tokenAddress: Address): Token {
...
-    token.decimals = 18;
+    const stablecoin = getStablecoinSymbol(network, tokenAddress);
+    if (stablecoin !== null) {
+      // Phase 2b stablecoins (USDC / USDC.e / pUSD) — all 6 decimals.
+      token.symbol = stablecoin;
+      token.decimals = 6;
+    } else {
+      // Fallback for any unknown token that shows up via classifyTransfer
+      // (shouldn't happen given our data-source set, but defensive).
+      token.symbol = "UNKNOWN";
```
DIvyaNautiyal07, line 643: **Should fix:** The UNKNOWN branch silently writes `decimals=18` with no log or signal. If a token is added to `networks.json` `erc20Tokens` but missed in `getStablecoinSymbol`, every Transfer row gets a `Token` entity with `symbol="UNKNOWN", decimals=18`. Because `getOrCreateToken` early-returns on `token != null` (line 610), the first write wins forever — any frontend that uses `Token.decimals` to format amounts is silently off by 10^12 for a 6-decimal stablecoin. Add at minimum a `log.critical` (or `log.warning`) naming the address and network.

subgraphs/pearl-transactions/tests/phase-2a.test.ts
```diff
@@ -694,4 +710,52 @@ describe("pearl-transactions / Phase 2a — raw OLAS + Safe template", () => {
       "MASTER_FUNDING_IN"
     );
   });
+
+  // ---- Phase 2b — stablecoins (USDC / USDC.e / pUSD, all 6 decimals) ----
```
DIvyaNautiyal07: **Should fix:** `getStablecoinSymbol` has 8 distinct (chain, address) cases — the two new tests cover 2 of them (Gnosis USDC.e, Polygon pUSD). The 4 canonical USDC entries (the highest-volume stablecoin on every chain) and the entire Optimism branch are exercised by zero tests. A one-character typo in any USDC address literal in `constants.ts` or in the Optimism block (lines 147-150) would silently fall through to UNKNOWN / 18-decimals. Add at minimum a Gnosis USDC test, a Base USDC test (Base is single-stablecoin so it's untested as a whole chain), and one Optimism test.

subgraphs/pearl-transactions/tests/phase-2a.test.ts
```diff
@@ -694,4 +710,52 @@ describe("pearl-transactions / Phase 2a — raw OLAS + Safe template", () => {
...
+      id.toHexString(),
+      "token",
+      USDCE_GNOSIS.toHexString()
+    );
+    assert.fieldEquals("Token", USDCE_GNOSIS.toHexString(), "symbol", "USDC.e");
+    assert.fieldEquals("Token", USDCE_GNOSIS.toHexString(), "decimals", "6");
+  });
+
+  test("Polygon pUSD Transfer resolves to pUSD @ 6 decimals", () => {
+    dataSourceMock.setNetwork("matic");
```
DIvyaNautiyal07: **Should fix:** This calls `dataSourceMock.setNetwork("matic")` mid-test and never resets back to `gnosis`. `afterEach` only does `clearStore()`. It's currently the last test in the file so no leak today, but any test appended after this one will silently inherit `matic` instead of the `beforeEach` promise of `gnosis`. Add `dataSourceMock.setNetwork("gnosis")` at end-of-test, or move the matic test into its own describe block with a proper beforeEach.

subgraphs/pearl-transactions/src/utils.ts
```diff
@@ -617,10 +618,17 @@ export function getOrCreateToken(tokenAddress: Address): Token {
     token.symbol = getWrappedNativeSymbol(network);
     token.decimals = 18;
   } else {
-    // Fallback for any unknown token that shows up via classifyTransfer
-    // (shouldn't happen given our data-source set, but defensive).
-    token.symbol = "UNKNOWN";
-    token.decimals = 18;
+    const stablecoin = getStablecoinSymbol(network, tokenAddress);
+    if (stablecoin !== null) {
+      // Phase 2b stablecoins (USDC / USDC.e / pUSD) — all 6 decimals.
```
DIvyaNautiyal07: **Nice to have (two comments):**

1. This inline branch comment `Phase 2b stablecoins (USDC / USDC.e / pUSD) — all 6 decimals` implies all three tokens exist on every chain. Only Polygon has all three (Gnosis / Optimism have two, Base has only USDC). Drop the parenthetical token list or qualify with `per chain`.

2. The function-level block comment a few lines above (around line 605) still reads `Phase 2b will extend with USDC (6 decimals) and USDC.e (6 decimals)` as a forward-looking TODO. This PR IS Phase 2b and the function already does it — and the comment also omits pUSD. Update to describe what the function does now.

subgraphs/pearl-transactions/src/constants.ts
```diff
@@ -126,6 +126,33 @@ export function getWrappedNativeSymbol(network: string): string {
   return "WNATIVE";
 }
 
+// getStablecoinSymbol — Phase 2b stablecoin metadata resolver. Returns
+// the display symbol for a known per-network stablecoin (all 6 decimals),
+// or null if `address` isn't a tracked stablecoin. Addresses mirror the
+// `erc20Tokens` set in networks.json (sourced from the Operate app's
+// frontend/config/tokens.ts). All comparisons are lowercase-hex
+// (graph-ts toHexString() is lowercase).
+export function getStablecoinSymbol(
```
DIvyaNautiyal07, line 135: **Nice to have:** The stablecoin address set lives in three places: `networks.json` `erc20Tokens` (drives indexed events), this function (drives `Token.symbol` / `decimals`), and the generated YAML manifests. networks.json → YAML is mechanical, but networks.json → this resolver is hand-edited. A token added to one and missed in the other silently corrupts `Token` rows (see also the UNKNOWN-fallback comment on `utils.ts`). Consider moving symbol / decimals into `networks.json` itself and generating this resolver — single source of truth.

subgraphs/pearl-transactions/src/constants.ts
```diff
@@ -126,6 +126,33 @@ export function getWrappedNativeSymbol(network: string): string {
...
+  network: string,
+  address: Address
+): string | null {
+  const a = address.toHexString();
```
DIvyaNautiyal07: **Nice to have:** The function depends on graph-ts `Address.toHexString()` returning lowercase — documented in the comment above. Adding `.toLowerCase()` explicitly costs nothing and removes a hidden dependency on graph-ts internal behavior: `const a = address.toHexString().toLowerCase();`.

subgraphs/pearl-transactions/src/constants.ts
```diff
@@ -126,6 +126,33 @@ export function getWrappedNativeSymbol(network: string): string {
...
+  if (network == "gnosis" || network == "xdai") {
```
DIvyaNautiyal07, line 142: **Nice to have:** This function accepts the aliases `xdai` and `polygon` for the same chains, but the sibling `getWrappedNativeSymbol` just above (line 122) only matches the short forms `gnosis` and `matic`. Harmless today (Studio uses short forms only) but inconsistent. Pick one convention for both.

subgraphs/pearl-transactions/tests/phase-2a.test.ts
```diff
@@ -694,4 +710,52 @@ describe("pearl-transactions / Phase 2a — raw OLAS + Safe template", () => {
       "MASTER_FUNDING_IN"
     );
   });
+
+  // ---- Phase 2b — stablecoins (USDC / USDC.e / pUSD, all 6 decimals) ----
+
+  test("Gnosis USDC.e Transfer routes via the generic handler; Token = USDC.e @ 6 decimals", () => {
```
DIvyaNautiyal07: **Nice to have:** These Phase 2b tests sit inside the `describe("pearl-transactions / Phase 2a — raw OLAS + Safe template", ...)` block from line 282. Anyone filtering test runs by describe name will see Phase 2b tests reported under Phase 2a. Either move into a new `describe("Phase 2b — stablecoins", ...)` block or rename the outer describe.

kupermind:
Reviewed end-to-end. Verifying technical claims independently and adding findings beyond what @DIvyaNautiyal07 already flagged in their inline pass. CI green ✅; **38/38 Matchstick passing**.

## ✅ Independently verified claims

### 1. Token set matches `valory-xyz/olas-operate-app` `frontend/config/tokens.ts` byte-for-byte

Cross-referenced all 8 (chain, token) tuples against [`olas-operate-app:frontend/config/tokens.ts`](https://github.com/valory-xyz/olas-operate-app/blob/main/frontend/config/tokens.ts):

| Chain | Token | Address | Upstream | Decimals |
|---|---|---|---|---|
| gnosis | USDC | `0xDDAfbb50…` | L120-125 ✅ | 6 ✅ |
| gnosis | USDC.e | `0x2a22f9c3…` | L114-119 ✅ | 6 ✅ |
| matic | USDC | `0x3c499c54…` | L213-218 ✅ | 6 ✅ |
| matic | USDC.e | `0x2791Bca1…` | L219-224 ✅ | 6 ✅ |
| matic | pUSD | `0xC011a7E1…` | L230-235 ✅ | 6 ✅ |
| optimism | USDC | `0x0b2C639c…` | L188-193 ✅ | 6 ✅ |
| optimism | USDC.e | `0x7F5c764c…` | L199-204 ✅ | 6 ✅ |
| base | USDC | `0x833589fC…` | L146-151 ✅ | 6 ✅ |

Upstream JSDoc explicitly warns on pUSD: *"pUSD has 6 decimals, not 18"*. Decimal classification is correct.

Sanity note on Gnosis USDC `0xDDAfbb50…`: this is Circle's native bridged USDC on Gnosis (deployed Nov 2024); the older Omnibridge variant is `0x2a22f9c3…` (USDC.e). Both legitimately tracked. ✓

Also confirmed: Mode-chain USDC (`0xd988097f…`) is in upstream but correctly excluded — Mode is deprecated per plan §3.3. ✓

### 2. `generate-manifests.js` change is a true no-op for the other 4 templated subgraphs

`grep -rn "erc20TokenDataSources\|erc20Tokens" subgraphs/ scripts/` returns matches only in pearl-transactions. And Python-scanning every other templated subgraph's `networks.json` for array-valued keys confirms **none** of staking / tokenomics-l2 / liquidity-l2 / service-registry hold any arrays at the per-network level:

```
--- staking ---           gnosis/base/optimism/mainnet/matic/arbitrum-one/celo: []
--- tokenomics-l2 ---     base/arbitrum-one/gnosis/optimism/matic/celo:        []
--- liquidity-l2 ---      gnosis/matic/arbitrum-one/optimism:                  []
--- service-registry ---  mainnet/optimism/base/celo/gnosis/matic/arbitrum-one: []
```

→ The new `if (Array.isArray(contractData)) continue;` guard never fires for them, and the `{{ erc20TokenDataSources }}` marker doesn't appear in any other template. Byte-identical output is mechanically guaranteed without re-running the script. ✓ (Author's pre-flight diff confirmation accepted.)

## 🟡 Concurring with @DIvyaNautiyal07's "Should fix" items

All three of their must-fix calls are real. Light enrichment:

- **UNKNOWN-fallback silent bad-decimals (utils.ts:629)** — agree. Compounding hazard: since `getOrCreateToken` is called by every classified Transfer, and graph-node *creates entities in WASM in the same block they're queried*, the first write of a 6-decimal stablecoin under `decimals=18` will persist forever (FundsMovement.amount stays correct — that's wei — but consumer-side formatting silently divides by `10^18`, hiding `1,000,000` USDC.e units as `0.000001 USDC.e`). At minimum add the `log.critical` they suggested. Stronger fix: **CI-time guard** — a tiny Node.js script run in CI that loads `networks.json` and greps `getStablecoinSymbol` to ensure every `erc20Tokens` entry has a matching resolver branch. Same script would catch the three-source-of-truth fragility @DIvyaNautiyal07 flagged for `constants.ts:135`.
- **Test coverage gap (test.ts:714)** — agree. 2 of 8 tuples covered. Worth highlighting: **no Optimism test exists at all** in the new suite, and Optimism is one of two chains where the resolver is the only thing distinguishing canonical USDC from USDC.e at the `Token.decimals` layer. A typo in either Optimism literal silently maps to `UNKNOWN @ 18`.
- **Test isolation leak (test.ts:740)** — agree. `dataSourceMock.setNetwork("matic")` without a reset is a time bomb. If you adopt @DIvyaNautiyal07's "move into its own describe block" suggestion, that block's `beforeEach` should explicitly `setNetwork("matic")` so the dependency is local, not assumed.

## 🔴 New findings (not in the inline review)

### A. §6.3 benchmark gate is being skipped, not just deferred

The PR description acknowledges this: *"§6.3a Polygon sync benchmark… deferred because Studio deploy isn't provisioned yet. On-chain path chosen per product."*

But plan §6.3 is unambiguous (lines 896–910):

> *"Before any commitment, run a benchmark… If not acceptable → do not ship 2b on-chain."*

§2.2 explicitly tags Polygon USDC.e as **the** cost hotspot, with the §6.3 benchmark gate existing *specifically* to manage it. Landing the on-chain path without that measurement turns the mitigation *"revisit if Polygon full-sync proves too slow"* into post-hoc remediation — by the time it's noticed, the deploy is provisioned, dashboards are wired, and rolling back to off-chain stablecoins is a wallet-UI refactor, not a config change.

This isn't a "should fix in code" — it's a process question. Recommend at minimum:
- (a) Add a 1-sentence Studio dashboard call-out (sync rate target, e.g. "if Polygon sync rate < X events/sec, rip out Polygon USDC.e and document the off-chain fallback") so the rollback condition is named, not implicit.
- (b) Land Polygon USDC.e behind a separate manifest entry (it's the §2.2 cost hotspot) so it can be removed without re-rendering the rest of the matrix if the benchmark fails post-deploy.
- (c) **Alternately,** carve Polygon USDC.e out of this PR and ship gnosis/optimism/base + Polygon canonical USDC + pUSD now, leaving the actual cost hotspot for the benchmark conversation. The 1-line change in `networks.json` (drop the USDC.e entry on `matic`) lets the same plumbing test on a low-risk slice first.

### B. Plan §4.5 / §6.3 token list is now stale vs. shipped code

PR description acknowledges: *"plan §6.3 assumed Polystrat funds in USDC.e on Polygon, but the current app config shows it uses USDC + pUSD"*. Plan §4.5 (the asset inventory table) doesn't list pUSD at all — only USDC and USDC.e on Polygon. This adds **another** concrete plan-vs-code divergence on top of the 14 already raised on PR #129 ([comment 4593942721](https://github.com/valory-xyz/autonolas-subgraph-studio/pull/129#issuecomment-4593942721)).

Two options:
- Tack a §4.5 token-table update + §6.3 polystrat-currency correction onto the Rev. 6 sweep on PR #129 *before* it merges.
- Or, since this PR also ships outside the plan, push a small follow-up plan PR after #129 lands.

Either way, flagging so it doesn't fall through the cracks.

### C. Several `erc20Tokens` `startBlock` values are wasted indexing

`startBlock` is set to the chain's `ServiceRegistryL2` deploy (provably-safe lower bound, fine in principle). But the actual token contracts deployed much later on several chains:

- **Gnosis USDC `0xDDAfbb50…`** — Circle native, deployed ~Nov 2024. SRL2 startBlock 27,871,084 ≈ Jul 2023. **~16 months of empty block scanning** before the contract exists.
- **Polygon canonical USDC `0x3c499c54…`** — deployed Nov 2023. SRL2 startBlock 41,783,952 ≈ Apr 2023. ~7 months of empty scanning.
- **Polygon pUSD `0xC011a7E1…`** — Polymarket USD, deployed approx. Q1 2024. Several months of empty scanning.

This isn't wrong (event-decoding cost on empty blocks is small relative to actual indexing), but combined with point A above it's a cheap incremental win: per-token deploy blocks would shave a sync chunk off each, and `getStablecoinSymbol`'s 6-decimal claim already implicitly assumes these are the canonical Circle/Polymarket deploys, so the deploy-block lookup is unambiguous. If you reopen `networks.json`, consider `{ "name": "USDC", "address": "0x…", "startBlock": … }` per token rather than re-using `ServiceRegistryL2.startBlock`.

### D. Operational hygiene — `erc20Tokens` array order

The generated manifest's data-source order follows the `erc20Tokens` array order. graph-node is mostly indifferent, but reordering existing entries between deploys can disturb grafting/partial-reindex behavior. Worth a one-liner in `networks.json` header comment: *"treat `erc20Tokens` as append-only — reorder only via a fresh deploy"*. Tiny doc nit but cheap.

## 🟡 @DIvyaNautiyal07's "Nice to have" items — concur all

All five are reasonable. Two are essentially free:
- `.toLowerCase()` defensiveness in `getStablecoinSymbol` (one method call, removes a hidden dep on graph-ts internals)
- Move the two new tests into a `describe("Phase 2b — stablecoins", ...)` block (re-categorization, no logic change)

The "single source of truth for `(address → symbol, decimals)`" suggestion is the strongest cleanup of the bunch — see point A above for the CI-time guard alternative if generating the resolver from `networks.json` proves too invasive.

## Summary

- Token set ✅ verified upstream; decimals ✅ correct; generate-manifests change ✅ verified no-op for other subgraphs.
- @DIvyaNautiyal07's 3 "Should fix" + 5 "Nice to have" items: all valid, concur.
- **New blocking-ish concern:** §6.3 benchmark gate skipped — recommend at minimum carving Polygon USDC.e into its own follow-up (point A.c).
- **Process concerns:** plan-vs-code drift now widens (point B); will need to roll into the Rev. 6 sweep on #129.
- Deploy-block / array-order are minor operational hygiene (points C, D).

Holding off on a merge recommendation until the §6.3 benchmark question is resolved with the team.

rajat2502:
Thanks @DIvyaNautiyal07 + @kupermind — pushed `8a5ee33` addressing the code items. 44/44 Matchstick passing.

## ✅ "Should fix" (all done)
- **UNKNOWN-fallback (utils.ts:629)** — now `log.critical`s the token address + network in the fallback branch, and the comment explains the first-write-wins / 10^12-misformat hazard. (Went with the loud runtime guard; see note on the CI-time guard below.)
- **Test coverage (test.ts:714)** — now covers **all 8 (chain, token) tuples** (was 2), including the entire previously-untested **Optimism** branch and all four canonical USDC entries. A typo in any address literal now fails here instead of silently mapping to UNKNOWN/18.
- **Test isolation (test.ts:740)** — Phase 2b cases moved into their own `describe("Phase 2b — stablecoins")` block with a `beforeEach`/`afterEach` that pins **and resets** the network. No more `setNetwork("matic")` leak.

## ✅ "Nice to have"
- `.toLowerCase()` added to `getStablecoinSymbol` — drops the graph-ts casing dependency.
- Phase 2b tests no longer reported under the Phase 2a describe (own block now).
- Stale comments on `getOrCreateToken` updated to describe current behaviour (+ dropped the misleading "all three tokens" parenthetical — only Polygon has all three).
- **Alias consistency (constants.ts:140)** — re-checked: `getWrappedNativeSymbol` *already* handles `xdai`/`polygon` (`network == "gnosis" || network == "xdai"`, etc.), as do all the other resolvers. The code is already consistent; no change needed. (Possibly read against an older state.)
- **Single-source-of-truth / CI-time guard** — deferred, not ignored. Generating the resolver from `networks.json` (or a CI script asserting every `erc20Tokens` entry has a resolver branch) touches shared CI/codegen infra, so I'd rather do it as its own follow-up. In the meantime the **runtime `log.critical` + full 8-tuple test coverage** together close the actual drift risk (a mismatch now either fails a test or screams in logs). Filed as a follow-up; happy to do it if you'd prefer it in this PR.

## Finding D (append-only array order)
Documented in `generate-manifests.js` (networks.json is strict JSON — can't take a comment, and a `"//"` key would break the generate loop). Noted erc20Tokens is append-only + the resolver-sync requirement.

## Needs a decision — not code (findings A, B, C)

**A — §6.3 benchmark gate (the merge-blocker).** Agreed this is a process call, not code. The product decision was "ship on-chain"; the benchmark was deferred only because **Studio isn't provisioned** (nothing's deployed yet — there's no environment to benchmark in). Your option (c) — carving Polygon USDC.e (the §2.2 hotspot) into a separate follow-up so the rest ships now and the hotspot waits for the benchmark — is clean and I can do it as a 1-line `networks.json` change. **@<product> — do you want (c) carve-out, or ship all of 2b on-chain now with a named rollback condition (option a)?** Holding the merge on this answer per @kupermind.

**B — plan §4.5 / §6.3 drift (pUSD missing, Polystrat currency).** Real, and it's part of the same plan-vs-code reconciliation as the #129 Rev. 6 sweep. I'll fold the §4.5 token-table + §6.3 currency fix into #129 before it merges (that PR is still open) rather than a separate plan PR.

**C — per-token startBlocks.** Optional optimization (empty-block scanning before each token's deploy). Tied to whether we reopen `networks.json` for A anyway. If we do the (c) carve-out, I'll add per-token `startBlock`s in the same edit; otherwise it's a minor follow-up.

kupermind:
Re-reviewed against your fix commit `8a5ee33` + merge of latest `main`. **All "Should fix" + addressable "Nice to have" items cleanly addressed.** CI green ✅ (44/44 Matchstick passing — up from 38, +6 from the new 8-tuple coverage). One process loose end on Finding B I should flag.

## ✅ "Should fix" — all done correctly

**1. UNKNOWN-fallback `log.critical` (utils.ts:629)** — added with a great error message:
```
"getOrCreateToken: no symbol/decimals resolver for indexed token {} on {} —
 add a getStablecoinSymbol branch (networks.json `erc20Tokens` and the resolver
 are out of sync). Falling back to UNKNOWN/18."
```
Names the root cause + the specific files to touch + the fallback behavior — operator who hits this in indexing logs has everything they need to fix it without context-switching. The rewritten function comment now also explicitly calls out *first-write-wins makes wrong decimals permanent — hence the `log.critical` guard*, so future maintainers know why it's `critical` rather than `warning`. ✅

**2. Test coverage 2/8 → 8/8 (tests/phase-2a.test.ts:714)** — done via a clean `assertStablecoinResolves(network, address, expectedSymbol, salt)` helper that consolidates the seed + transfer + assertion pattern. All 8 (chain, token) tuples covered — including the entire previously-untested Optimism branch (`gnosis USDC`, `gnosis USDC.e`, `matic USDC`, `matic USDC.e`, `matic pUSD`, **`optimism USDC`**, **`optimism USDC.e`**, **`base USDC`**). A typo in any address literal in `constants.ts` now fails a specific test rather than mapping to UNKNOWN/18. ✅

Nice touch: the helper is intentionally file-scoped (not nested), with a comment explaining it's *"to avoid the AssemblyScript closure-capture WASM-validator error"* — saves the next person from re-discovering that AS gotcha.

**3. Test isolation (tests/phase-2a.test.ts:740)** — Phase 2b cases moved into their own `describe("pearl-transactions / Phase 2b — stablecoins", ...)` block with `beforeEach`/`afterEach` that both pin **and reset** `dataSourceMock.setNetwork("gnosis")`. The `setNetwork("matic")` leak is gone — even if a future test were appended to a different block, the reset means it can't inherit network state from this suite. ✅

## ✅ "Nice to have" — all addressed (the legitimate ones)

- ✅ `.toLowerCase()` in `getStablecoinSymbol` with a comment explaining why (drops the graph-ts casing dependency). Free defensiveness.
- ✅ Phase 2b tests are in their own describe block (no longer reported under Phase 2a).
- ✅ `getOrCreateToken` function comment rewritten to describe current behaviour (was a stale forward-looking TODO) — includes the `first write wins` warning right where the next maintainer touches the code.
- ✅ Misleading "all three tokens" parenthetical dropped (only Polygon has all three).
- ✅ Append-only convention for `erc20Tokens` documented in `generate-manifests.js` (smart placement — the file can't take a comment in JSON itself, and the script is the only thing that consumes the array).

**Alias-consistency pushback (constants.ts:140) — verified correct.** Spot-checked at `subgraphs/pearl-transactions/src/constants.ts:122-128`:

```ts
export function getWrappedNativeSymbol(network: string): string {
  if (network == "gnosis" || network == "xdai") return "WXDAI";
  if (network == "matic" || network == "polygon") return "WPOL";
  if (network == "optimism" || network == "base") return "WETH";
  return "WNATIVE";
}
```

`xdai`/`polygon` aliases were already there — the original review comment was off. ✅ No change needed; thanks for double-checking instead of just doing it.

**Single-source-of-truth / CI-time guard — deferred is fine.** Agree this is its own follow-up. Between the **runtime `log.critical`** (loud in indexing logs) + **8/8 test coverage** (compile-time-ish catch on any literal typo), the actual drift risk is closed. The CI-time guard would still be a nice-to-have but it's no longer carrying load — happy with the deferral.

## Findings A / B / C status

**Finding A — §6.3 benchmark gate.** Still awaiting product. Your suggested options are clean: (a) ship-with-named-rollback-condition or (c) carve out Polygon USDC.e into a follow-up. Both are reasonable — (c) is the safer move (Polygon USDC.e is *the* §2.2 cost hotspot; surfacing it on its own is the fastest rollback if it overwhelms sync). No code change needed in this PR until product decides.

**Finding B — plan §4.5 / §6.3 drift.** ⚠️ **This is now stuck.** You said in [your 17:14 comment](https://github.com/valory-xyz/autonolas-subgraph-studio/pull/138#issuecomment-...):

> *"I'll fold the §4.5 token-table + §6.3 currency fix into #129 before it merges (that PR is still open)."*

But PR #129 merged at **16:52:52** — ~22 minutes *before* your comment. So the §4.5 fold-in didn't land in #129's Rev. 6 sweep, and the plan as it stands on `main` still says:

- §4.5: *"USDC.e (a.k.a. pUSD in Pearl UI)"* — single Polygon stablecoin row tagged as both. But shipped #138 treats **pUSD `0xC011a7E1…`** as a *separate* token from USDC.e `0x2791…`. Three separate Polygon stablecoins exist in the code, only one row in the plan.
- §6.3: *"polystrat funding flows in USDC.e historically"* — but per the PR description here, the current app config shows polystrat now uses USDC + pUSD.

Net: this PR's merge will re-introduce plan-vs-code drift in §4.5 + §6.3. Either:
- (a) Add a small commit to this PR doing the §4.5 + §6.3 fix as part of the same atomic unit (preferred — keeps plan-and-code-together discipline), or
- (b) Land #138, then immediately open a `docs(pearl-funds): Rev. 7 — Phase 2b token-set + polystrat currency` follow-up before any further code PRs.

(a) is much cleaner — the plan-edits are mechanical (replace the single "USDC.e / pUSD aliased" row with three rows; update the polystrat sentence). Maybe 20 lines.

**Finding C — per-token startBlocks.** Still optional; only worth the reopen if you reopen `networks.json` for B(a) or A(c) anyway.

## Status

LGTM on the fix commit. Only remaining items are:
1. ⚠️ **B** — close the plan-vs-code drift before/after this lands.
2. 🤝 **A** — product decision on §6.3 (full ship vs. Polygon-USDC.e carve-out).

Code itself: ready to merge once A is answered.

rajat2502:
Decisions from the team on the two open items:

**A — §6.3 benchmark gate → ship all of 2b on-chain now (no carve-out).** The benchmark can't run pre-deploy (Studio isn't provisioned), so we're shipping all three Polygon stablecoins on-chain and monitoring post-deploy. The **rollback condition is now named** in the plan (§6.3, Rev. 7): if Polygon USDC.e sync proves too slow once deployed, drop the single `matic` USDC.e `erc20Tokens` entry — a one-line `networks.json` change, no re-render of the rest. So #138 is good to merge as-is on this front; no code change here.

**B — plan §4.5/§6.3 drift → fixed in #139** (`docs(pearl-funds): Rev. 7`). Since #129 merged before this could fold in, it's a standalone plan PR. It reconciles the big one you caught — **pUSD is a separate Polymarket contract `0xC011a7E1…`, not a USDC.e alias** (the plan's "no separate contract exists" was stale; #138's three-token Polygon set is correct) — plus rebuilds the §4.5 table to the shipped per-chain set.

**C — per-token start blocks:** since we're not reopening `networks.json` for a carve-out, leaving this as the documented `ServiceRegistryL2`-block lower bound (minor empty-block scanning). Optional future tidy-up.

Net: code here is unchanged since your LGTM (`8a5ee33`); the two non-code items are resolved (A = ship + named rollback; B = #139). Thanks both for the thorough pass.

mohandast52 (approved):

DIvyaNautiyal07 (approved):

PR #139, opened 2026-06-02
docs(pearl-funds): Rev. 7 — reconcile Phase 2b stablecoin token set with shipped code

rajat2502:
## Summary

Closes the Phase-2b plan-vs-code drift @kupermind flagged on [#138](https://github.com/valory-xyz/autonolas-subgraph-studio/pull/138) — it couldn't be folded into the Rev. 6 sweep because #129 merged ~22 min before that comment landed. Docs-only.

## The core correction: pUSD is a separate token

The plan asserted (§4.5) that **pUSD = USDC.e on Polygon, "no separate token contract exists."** That's stale. The Operate app's `config/tokens.ts` — and the shipped #138 — list **pUSD `0xC011a7E1…` as a distinct Polymarket contract**, separate from bridged USDC.e `0x2791…`. So Polygon has **three** stablecoins (USDC, USDC.e, pUSD), not the one aliased row the plan showed.

## Changes

- **§4.5 token table** rebuilt to the shipped per-chain set. (Also fixes pre-existing gaps: USDC.e was listed Polygon-only but ships on gnosis/optimism too; USDC omitted Gnosis `0xDDAfbb50…`.)
- **§4.3 networks table** stablecoin column corrected to match.
- **§6.3** now records the decision: **shipped on-chain (#138), §6.3a benchmark deferred** (Studio not provisioned — no environment to measure in), with a named **Polygon-USDC.e rollback condition** (drop the one `matic` USDC.e `erc20Tokens` entry if its sync proves too slow). Polystrat funding corrected to **USDC + pUSD** (was: USDC.e).
- **§1.2 / §3** incidental `USDC.e(pUSD)` mentions corrected.

## Notes

- Plan and shipped code (#131/#132/#133/#138) are consistent again after this lands.
- The §6.3 benchmark is deferred, not skipped silently — the rollback condition is now named in the plan so "merge all and see" is actionable post-deploy.

Related: plan #129 (Rev. 1–6), phase-2b #138.


DIvyaNautiyal07 (dismissed):
Approving — all token addresses and per-chain rows are internally consistent between §4.3 and §4.5, and the §6.3 rewrite cleanly captures the ship-on-chain decision + the concrete Polygon-USDC.e rollback action. The §4.5 Notes block explicitly flagging the old `pUSD = USDC.e alias` claim as superseded is the right call — it leaves a paper trail so future readers don't re-litigate the Rev. 6→7 correction.

No inline comments — nothing actionable. Ready to merge.

kupermind:
Re-verified Rev. 7 against shipped code at `8a5ee33` (#138). The token-set diff is correct — all addresses match `networks.json` + `getStablecoinSymbol` byte-for-byte:

| Chain | Plan §4.5 | networks.json + constants.ts | OK |
|---|---|---|---|
| Gnosis | USDC `0xDDAfbb50…` + USDC.e `0x2a22f9c3…` | same | ✅ |
| Polygon | USDC `0x3c499c54…` + USDC.e `0x2791bca1…` + pUSD `0xC011a7E1…` | same | ✅ |
| Optimism | USDC `0x0b2C639c…` + USDC.e `0x7F5c764c…` | same | ✅ |
| Base | USDC `0x833589fC…` | same | ✅ |

§6.3's "shipped on-chain, benchmark deferred + Polygon USDC.e rollback condition" is the right call given Studio still isn't provisioned. Phase-table row updated. CI is the upstream graph-cli matchstick-binary fetch flake (`Invalid version. Must be a string. Got type "undefined"`), unrelated to docs.

**Two stale references the sweep missed — worth folding in before merge so Rev. 7 actually lands plan-vs-code clean:**

**1. `§4.3` intro (line 219).** Still says:
> "USDC / USDC.e (Phase 2b) is benchmark-gated per §6.3."

That contradicts the new §6.3 ("Shipped on-chain; benchmark deferred") and omits pUSD. Suggest:
> "USDC / USDC.e / pUSD (Phase 2b) is shipped on-chain (#138) per §6.3."

**2. `§4.7 / Section C` polystrat flow diagram (lines 482–510)** still bakes "USDC.e" into every label:

```
MEOA -->|"USDC.e funding..."| MSAFE
MSAFE -->|"polystrat capital (USDC.e)..."| ASAFE
MSAFE -->|"gas top-up (POL or USDC.e)..."| AEOA
ASAFE -->|"place bet (USDC.e collateral)"| POLY
```

…and the preamble says:
> "Where Pearl predict's *stablecoin* (USDC.e / pUSD; or canonical USDC post-migration) moves."

Per Rev. 7's own §4.5 + §6.3.b, **polystrat funds in USDC + pUSD** now, USDC.e is the §2.2 hotspot but not the canonical funding token. Suggest either (a) generic "stablecoin (USDC / pUSD)" arrow labels, or (b) explicit dual paths.

Both are cosmetic, but the §4.3 line in particular flatly contradicts §6.3's headline two screens down — exactly the kind of intra-doc drift Rev. 7 set out to kill.

**Verified clean:**
- Revision header summary is honest and self-consistent.
- All §4.5 addresses match shipped code (checked lowercase against `constants.ts:135-156`).
- §6.3 rollback condition (drop the `matic` USDC.e `erc20Tokens` entry) is mechanically a one-line change — `generate-manifests.js` skips arrays via `Array.isArray`, so the rollback won't disturb the rest of the matrix.
- "pUSD is a separate Polymarket contract" claim is correct (verified upstream Operate `frontend/config/tokens.ts` in #138 review last cycle).

Not blocking from my side — just two more red-pen passes and Rev. 7 is the first revision that's actually plan-vs-code clean across the whole document.

rajat2502:
Addressed both stale references in `b32c676`:
- **§4.3 intro** → "USDC / USDC.e / pUSD (Phase 2b) is shipped on-chain (#138) per §6.3" (was "benchmark-gated").
- **§4.7 / Section C polystrat flow** → funding arrows relabeled to USDC / pUSD per §4.5 / §6.3.b; the bet-collateral arrow stays USDC.e (correct — Polymarket collateral), and the preamble now calls out USDC.e as the §2.2 hotspot, not the funding token.

Thanks for the red-pen pass.

DIvyaNautiyal07 (approved):
Verified locally on the branch:

- All Polygon stablecoin addresses match `networks.json` and `getStablecoinSymbol` in `src/constants.ts` (USDC `0x3c499c54…`, USDC.e `0x2791bca1…`, pUSD `0xC011a7E1…`).
- Section C polystrat funding labels (USDC + pUSD) match §6.3.b and the current app-config story; the bet-collateral arrow staying USDC.e is correct.
- §6.3 rollback condition ("drop the `matic` USDC.e entry from `networks.json` `erc20Tokens`") maps cleanly to the actual `networks.json` shape — one entry to remove, no other coordination needed.

Two §6.3 ripple misses worth picking up in a follow-up (not blocking):

- Line 93 §2.2 still says "the benchmark gate in §6.3 exist specifically to manage this." §6.3 isn't a gate anymore; suggest "rollback condition in §6.3."
- Line 1208 §11 #2 still calls Phase 2b feasibility "the single biggest unknown." The same list strikes through resolved items (#1, #5, #6) with `~~old~~ **Resolved:** …` — apply the same pattern here pointing at the new §6.3 (shipped on-chain, benchmark deferred).

LGTM otherwise.

kupermind (approved):
Verified fix commit `b32c676` against the diff:

- ✅ **§4.3 intro (line 219)** — "USDC / USDC.e (Phase 2b) is benchmark-gated per §6.3" → "USDC / USDC.e / pUSD (Phase 2b) is shipped on-chain (#138) per §6.3". Matches §6.3's new headline.
- ✅ **§4.7 / Section C polystrat flow** — preamble rewritten to call USDC.e the §2.2 hotspot (not the funding token); funding arrows relabeled stablecoin / USDC / pUSD; bet-collateral arrow correctly kept as `USDC.e collateral` (that's the Polymarket-side asset, not Polystrat funding). All three funding arrows align with §4.5 / §6.3.b.

Divya's two `LGTM otherwise` follow-up items (§2.2 line 93 wording + §11 #2 line 1208 strike-through pattern) are deferrable — they're ripple wording rather than plan-vs-code drift. Fine to land Rev. 7 now and pick them up in a doc-polish follow-up.

LGTM — approving.

PR #115, opened 2026-04-21
feat(predict-polymarket): generalize from polystrat-only to cohort-agnostic

kupermind:
## Summary

Generalizes the existing `subgraphs/predict/predict-polymarket/` subgraph in place so that every Olas multisig on Polygon is indexed, with cohort filtering (polystrat, Pearl Mini, etc.) resolved client-side via a `multisig_:` link on `TraderAgent`. This unblocks Pearl Mini analytics without duplicating service-level data already in `service-registry/`, and fixes a structural issue (polystrat-only `agentId == 86` gate) in the running deployment.

> This PR went through four rounds of scope refinement as docs-only first, then shipped with the implementation once the design was agreed. See "What changed in response to review" at the bottom for the trail.

### Schema (`schema.graphql`)

- **Removed** `TraderService` (polystrat-only gate).
- **New** `Multisig` entity — minimal index of every Olas multisig on Polygon, keyed on the multisig address. Carries `serviceId`, `agentIds: [Int!]!`, `operators: [Bytes!]!` (all deduplicated), `traderAgent` back-link, `createdAt`, `terminatedAt`.
- **New** `multisig: Multisig!` link on `TraderAgent`.
- **New** `PayoutSource` enum + `source: PayoutSource!` on `PayoutRedemption` (`CONDITIONAL_TOKENS` vs `NEG_RISK_ADAPTER`).
- **Internal helpers** `ServiceIndex` (serviceId → multisig lookup) and `PendingMultisig` (buffers `RegisterInstance` events that fire before `CreateMultisigWithAgents`, the typical initial-deployment order on Polygon). Not part of the public contract; documented in `pearl-trades-schema.md` §3.4.

### Handlers (`src/`)

- **`service-registry-l-2.ts`** — dropped the `agentId == 86` gate. `handleRegisterInstance` appends `(agentId, operator)` to `Multisig` when it exists (looked up via `ServiceIndex`), else buffers on `PendingMultisig`. `handleCreateMultisigWithAgents` creates the `Multisig`, drains `PendingMultisig`, writes `ServiceIndex`. New `handleTerminateService` sets `Multisig.terminatedAt`.
- **`ctf-exchange.ts`** — `handleOrderFilled` loads `Multisig` by maker (early-return if absent), lazy-creates `TraderAgent` on first trade with `multisig: Multisig!` link, sets the `Multisig.traderAgent` back-link, and bumps `Global.totalTraderAgents` once per trading service.
[...]

PR #143, opened 2026-06-04
feat(predict-polymarket): index every Olas service; tag PayoutSource

kupermind:
## Summary

Drops the `agentId == 86` (polystrat-only) gate in `service-registry-l-2.ts:14`, so every Olas service registered on `ServiceRegistryL2` (Polygon) is indexed by this subgraph. Cohort filtering moves from a hardcoded constant to deduplicated arrays on `TraderService` — the same model `pearl-transactions:Service.agentIds / Service.operators` already uses — so consumers filter polystrat / Pearl Mini / future cohorts client-side without a cross-subgraph join.

Bundled: tag the new `PayoutSource` enum on `PayoutRedemption` rows so dashboards can distinguish the three on-chain redemption paths.

**Replaces #115** — that PR's Multisig entity + scoping docs were architecturally overtaken by the v2-exchange overlay (`e29e508`) and the shipped `pearl-transactions/` subgraph. This scoped-down change preserves only the still-valid bug fix.

### Schema (`schema.graphql`)

- `TraderService` → mutable; gains `agentIds: [Int!]!`, `operators: [Bytes!]!` (dedup arrays), `traderAgent: TraderAgent` (reverse link).
- `TraderAgent.traderService: TraderService!` (forward link).
- New `PayoutSource` enum (`CONDITIONAL_TOKENS` / `NEG_RISK_ADAPTER` / `COLLATERAL_ADAPTER`); new `PayoutRedemption.source: PayoutSource!`.

### Handlers

- `service-registry-l-2.ts` rewritten: drop the `PREDICT_AGENT_ID` gate; load-or-create `TraderService`, dedup-append `(agentId, operator)` per registration. Mirrors `pearl-transactions:appendServiceRegistration` / `bufferPendingRegistration` (same dedup semantics across both subgraphs). `handleCreateMultisigWithAgents` defensively load-or-creates `TraderService` (in case it fires first) and wires both forward + reverse links.
- `utils.ts:processRedemption` takes a `source: string` arg; stamps it on the `PayoutRedemption` entity.
- `conditional-tokens.ts` / `neg-risk-mapping.ts` / `collateral-adapter.ts` pass `CONDITIONAL_TOKENS` / `NEG_RISK_ADAPTER` / `COLLATERAL_ADAPTER` at each call site.

### `subgraph.yaml`

Declares `PayoutRedemption` on the two redemption data sources where it was missing (`ConditionalTokens`, `NegRiskAdapter`). The four `*CollateralAdapter` data sources already declared it.

### Tests (111 → 111 passing)

- `service-registry-l-2.test.ts` rewritten for the lifted gate: cohort rows for both polystrat (86) and a Pearl-Mini surrogate agent ID (87), dedup semantics, multi-agent accumulation, both link directions, defensive load-or-create when `CreateMultisigWithAgents` fires first.
- `payout-source.test.ts` **new**: asserts `PayoutRedemption.source` is tagged correctly for all three redemption paths.
- `profit.test.ts`, `ctf-exchange.test.ts`, `ctf-exchange-v2.test.ts`: seed a `TraderService` next to every directly-constructed `TraderAgent` (now required by the schema link).

### Docs (`predict-polymarket/CLAUDE.md`)

End-to-end refresh: drops the "agent ID 86 only" framing, adds a Cohort filtering section with both `agentIds_contains` and `operators_contains` patterns, references `pearl-transactions` as the source of Master/Agent Safe + funds-movement metadata (join on the multisig address).

## Cohort filtering — consumer query patterns

```graphql
# Polystrat (the previous implicit filter)
traderAgents(where: { traderService_: { agentIds_contains: [86] } }) { ... }

# Pearl Mini — services registered via the PolySafeCreator operator
traderAgents(
  where: { traderService_: { operators_contains: ["0xA749f605..."] } }
) { ... }

# All Olas services on Polygon, regardless of whether they've ever traded
traderServices { id agentIds operators traderAgent { id totalBets } }
```

`PREDICT_AGENT_ID = 86` is retained in `constants.ts` purely as a client-side reference constant (no longer used as a gate).

## What's NOT in this PR

Carried over from PR #115's *Not in scope* — none of these are needed for the bug fix and they belong elsewhere:

- **`ApplicationClassifier`** — contract not deployed; deferred.
- **Full `AgentInstance` registration tuples** — belong in `subgraphs/service-registry/` or `pearl-transactions/`, not duplicated here.
- **Funding flows (USDC/MATIC `Transfer` indexing)** — covered by `pearl-transactions/`; off-chain reconstruction is viable for everything else.
- **Server-side joins** — hard policy boundary.

## Test plan

- [x] `yarn codegen && yarn build` clean on `subgraphs/predict/predict-polymarket/`.
- [x] `yarn test` — 111/111 Matchstick tests pass (locally).
- [x] `yarn audit:prod` — OK; `yarn audit:install-hooks` — OK; root `lockfile-lint` — clean.
- [ ] CI green on this PR.
- [ ] Sanity-check on a parallel deployment before cutover: confirm `traderAgents(where: { traderService_: { agentIds_contains: [86] } })` on the new deployment matches `traderAgents` (un-filtered) on the existing polystrat-only deployment for a representative window.
- [ ] Spot-check a Pearl-Mini multisig that registered post-launch — confirm it now appears in the new deployment but not the old.


rajat2502 (dismissed):
Approved. Lifts the `agentId == 86` gate so every Olas service on Polygon is indexed, with cohort identity moved onto deduplicated `agentIds` / `operators` arrays (mirroring the pearl-transactions model), plus a `PayoutSource` enum on `PayoutRedemption`.

Verified:
- `processRedemption` signature change covered at all 3 call sites.
- Required link `TraderAgent.traderService!` satisfiable — single `TraderAgent` construction site, service load-or-created first.
- Path-A/Path-B double-counting guard still holds (gate lift doesn't make adapter addresses TraderAgents).
- Test coverage strengthened, not weakened.

Consumer-facing notes (non-blocking): `Global.totalTraderAgents` now counts all Olas service multisigs (not just polystrat), and un-filtered `traderAgents` queries now return all cohorts — consumers relying on the implicit agent-86 filter must add `traderService_: { agentIds_contains: [86] }`. Both are in the PR test plan.

subgraphs/predict/predict-polymarket/src/service-registry-l-2.ts
```diff
@@ -44,6 +110,9 @@ export function handleCreateMultisigWithAgents(
 
     traderAgent.save();
 
+    service.traderAgent = traderAgent.id;
```
DIvyaNautiyal07: **[P2]** `service.traderAgent = traderAgent.id` is inside the `if (traderAgent === null)` block, so the reverse link is only written on the first sighting of the multisig.

On Olas the canonical scenario (one service → one new multisig) is fine. But if a multisig is ever observed via a second `CreateMultisigWithAgents` for a different `serviceId` (uncommon but technically reachable via Safe address reuse), the second service's defensive `TraderService` row stays with `traderAgent = null` forever, while the existing `TraderAgent.traderService` still points at the first service. Any client doing `traderService { traderAgent { ... } }` for the second service gets null and silently loses the agent's P&L.

Fix is one idempotent block after the `if`:

```typescript
if (service.traderAgent == null) {
  service.traderAgent = traderAgent.id;
  service.save();
}
```

**[Nit, same area]** The defensive load-or-create at lines 88–93 duplicates the service-creation half of `recordRegistration`. A small `ensureTraderService(serviceId): TraderService` helper would let both call-sites share it. Low value, just flagging.

subgraphs/predict/predict-polymarket/tests/ctf-exchange.test.ts
```diff
@@ -52,8 +52,18 @@ function createOrderFilledEvent(
 }
 
 function setupTraderAgent(address: Address, serviceId: BigInt): void {
```
DIvyaNautiyal07: **[P2]** This `setupTraderAgent` function is duplicated identically in `tests/ctf-exchange-v2.test.ts:66` — same body, same fields, same load-or-create. Both files already import from `tests/test-helpers.ts`, so moving the helper there and importing it would close the copy-paste maintenance hazard: if `TraderAgent` ever grows another required field, today both files need the same fix.

Not blocking, just cleanup.

subgraphs/predict/predict-polymarket/tests/service-registry-l-2.test.ts
```diff
@@ -38,175 +42,159 @@ function createCreateMultisigWithAgentsEvent(
...
+describe("ServiceRegistryL2 - handleRegisterInstance (gate lifted)", () => {
   beforeEach(() => {
     clearStore();
   });
 
-  test("Should create TraderService when agent ID matches PREDICT_AGENT_ID", () => {
-    let event = createRegisterInstanceEvent(OPERATOR, SERVICE_ID_1, AGENT_INSTANCE, BigInt.fromI32(PREDICT_AGENT_ID));
-
-    handleRegisterInstance(event);
+  test("Creates TraderService for the polystrat agentId (86)", () => {
```
DIvyaNautiyal07: **[Nit]** Test name reads "Creates TraderService for the polystrat agentId (86)" — but the point of the test post-PR is that **any** agentId is accepted, and 86 is just an example value being stored in the array. A future reader scanning the test names could think 86 is still a meaningful gate value.

Suggest `Creates TraderService recording an example agentId (86)` or `Creates TraderService for an arbitrary agentId`.

subgraphs/predict/predict-polymarket/schema.graphql
```diff
@@ -122,13 +133,26 @@ type Global @entity (immutable: false) {
   totalMarketsParticipated: Int! # count of unique markets where any agent participated
 }
 
+# Which on-chain path emitted the redemption event. CONDITIONAL_TOKENS and
+# NEG_RISK_ADAPTER are the pre-v2-cutover direct paths; COLLATERAL_ADAPTER
+# covers post-cutover redemptions routed through CtfCollateralAdapter and
+# NegRiskCtfCollateralAdapter (the standard and neg-risk wrappers share
+# the same `PositionsRedeemed` event signature). Lets downstream tools
+# distinguish the routing without re-deriving it from the tx call graph.
+enum PayoutSource {
```
DIvyaNautiyal07, line 156: **[Nit]** The enum comment correctly distinguishes the three paths, but doesn't note that post-v2-cutover `CONDITIONAL_TOKENS` and `NEG_RISK_ADAPTER` rows are effectively historical: post-cutover the legacy events fire with `redeemer = adapter address`, which trips the un-tracked-redeemer guard in `processRedemption` and the row never gets created. Future API consumers looking at production data may wonder why those two values rarely appear.

One-line addendum on the enum comment would explain it.

DIvyaNautiyal07 (commented):
Verified locally on the branch (cloned at /tmp/ass-pr143):

- Test count 113 — handler-to-source mapping for all three PayoutSource values covered. Lifted-gate regression test at `tests/service-registry-l-2.test.ts:66-75` would fail if the agentId gate were re-introduced.
- All 4 stablecoin call-sites pass the right PayoutSource constant; `constants.ts` strings match the schema enum exactly.
- AssemblyScript array re-assign in `recordRegistration` is correct (`service.agentIds = agentIds` after push) — the silent-no-op trap from in-place `.push()` is avoided.
- `subgraph.yaml` entity declarations match what each data source's handlers actually write via `processRedemption`.

Three findings outside the diff hunks (couldn't inline-anchor them):

**[P1] `subgraphs/predict/predict-polymarket/README.md:9`** still says "Tracks agents registered via ServiceRegistryL2 (agent ID 86)". The README is the public entry point; the PR rewrote claude.md but missed the README.

**[P1] `subgraphs/predict/predict-polymarket/claude.md:775`** under Project Structure still says `src/service-registry-l-2.ts: Agent registration (services with agent ID 86 only)`. Directly contradicts the rewritten sections at lines 56, 372, 943 of the same file. In-file ripple miss.

**[P1] `subgraphs/predict/predict-polymarket/src/utils.ts:319-327`** — `processRedemption` has three silent early returns. The line-319 `agent == null` return is intentional (post-cutover the legacy event fires with `redeemer = adapter`, this is the Path A / Path B double-count guard — worth a one-line comment on that line). But lines 323 (`question == null`) and 327 (`participant == null`) are not intentional filters — they signal a tracked agent redeeming from a market the subgraph never observed (start-block race, indexing gap), and when that happens `totalPayout` / `totalTradedSettled` / `Global` / `DailyProfitStatistic` all silently understate with no signal in indexing logs. Suggest `log.warning` on 323 and 327 tagged with `redeemer` / `conditionId` / `source`.

**[P2] `subgraphs/predict/predict-polymarket/schema.graphql:126`** — `totalTraderAgents` semantics shifted with the gate lift. Pre-PR it counted polystrat-only multisigs; post-PR it counts every Olas multisig on Polygon. Any external dashboard reading this as "polystrat headcount" silently gets inflated numbers. Either tighten the schema comment to make the new scope explicit, or rename. The cohort-specific equivalent is `traderAgents(where: { traderService_: { agentIds_contains: [86] } })`.

The rest follow inline. Pre-existing items (MarketMetadata.description query, totalPayout "xUSDC" comment, profit-formula inconsistency, TraderService.id vs pearl-transactions key format) are deliberately out of scope of this review.

kupermind:
Addressed Divya's review in `ec33661`. All P1 + P2 items in scope; nits + the PayoutSource enum addendum folded in. 114/114 Matchstick tests passing locally.

## P1 — file-internal ripple fixes

- **`README.md:9`** — "Tracks agents registered via ServiceRegistryL2 (agent ID 86)" → "Indexes every Olas service registered via ServiceRegistryL2 on Polygon; cohort filtering (polystrat agent ID 86, Pearl Mini operator, etc.) is client-side via `TraderService.agentIds` / `operators` (deduplicated arrays)."
- **`CLAUDE.md:775`** (Project Structure) — same rewrite on the `src/service-registry-l-2.ts` descriptor.

## P1 — `processRedemption` guard rationale + warnings

- **L319 `agent == null`** — kept as a silent return; added an inline comment naming the Path A / Path B double-count rationale (post-cutover the legacy events fire with `redeemer = adapter address` and this guard is what stops them double-counting against the `CtfCollateralAdapter` `PositionsRedeemed` handler).
- **L323 `question == null`** + **L327 `participant == null`** — added `log.warning` with `redeemer` / `conditionId` / `source`. The pre-existing silent returns were turning a tracked-agent-redeems-from-unindexed-market case into a silent `totalPayout` / `Global` / `DailyProfitStatistic` under-count.

## P2 — `Global.totalTraderAgents` semantic shift

`schema.graphql:126` — comment now spells out the new scope (every Olas multisig on Polygon, cross-cohort) and includes the polystrat-only equivalent query so dashboards don't silently inflate the headcount after re-reading:

```graphql
traderAgents(where: { traderService_: { agentIds_contains: [86] } })
```

## P2 — `service-registry-l-2.ts:113` idempotent reverse link

Fixed exactly as suggested — moved `service.traderAgent = traderAgent.id` outside the `if (traderAgent === null)` block, guarded by `existingLink === null`. New test (`Idempotent reverse link survives Safe address reuse across services`) asserts the invariant on a contrived two-service-shares-multisig flow.

Small AS-compiler hiccup: direct `service.traderAgent === null` on the `Bytes | null` accessor crashes the AssemblyScript compiler in this version; assigning to a local first (`const existingLink = service.traderAgent`) is the standard workaround. Same shape pearl-transactions uses for `service.masterSafe`-style nullable refs.

**Companion refactor:** extracted a small `ensureTraderService(serviceId)` helper from `recordRegistration` and reused it at the multisig-time load-or-create site, so both call sites share the same load-or-create.

## P2 — `setupTraderAgent` test-helper consolidation

Moved into `tests/test-helpers.ts`; both `ctf-exchange.test.ts` and `ctf-exchange-v2.test.ts` now import it. One place to update if `TraderAgent` ever grows another required field.

## Nits

- **`tests/service-registry-l-2.test.ts:55`** — renamed `Creates TraderService for the polystrat agentId (86)` → `Creates TraderService recording an arbitrary agentId (here, polystrat 86)` so future readers don't read 86 as a still-meaningful gate.
- **`schema.graphql` PayoutSource enum** — added a note that CT / NEG_RISK_ADAPTER are effectively historical post-cutover (their legacy events still fire but get dropped by the un-tracked-redeemer guard), and new rows are predominantly `COLLATERAL_ADAPTER`. Saves future API consumers from wondering why those two values rarely appear.

## Local validation

- `yarn codegen && yarn build` — clean.
- `yarn test` — 113 → **114 passing** (the new idempotent-link test).
- `yarn audit:prod` — OK; `yarn audit:install-hooks` — OK.

Pre-existing items Divya flagged as out-of-scope (`MarketMetadata.description`, `totalPayout` "xUSDC" comment, profit-formula inconsistency, `TraderService.id` vs pearl-transactions key format) deliberately left for a follow-up.

DIvyaNautiyal07 (approved):
All prior P1/P2 findings addressed in `ec33661`:

- README.md and claude.md rewritten to drop the "(agent ID 86)" framing and document client-side cohort filtering on `TraderService.agentIds` / `operators`.
- `utils.ts` processRedemption: `agent == null` got the Path A/Path B double-count rationale inline; `question == null` and `participant == null` now emit `log.warning` tagged with `conditionId` / `redeemer` / `source` so a tracked agent redeeming a market we never observed surfaces in indexing logs instead of silently under-counting `totalPayout` / `Global`.
- `schema.graphql:126` comment clarifies the new cross-cohort scope of `totalTraderAgents` and points to the polystrat-only equivalent query.

Plus useful extras: PayoutSource enum post-cutover note, an idempotent reverse-link fix for the Safe-address-reuse edge (with a new test pinning it), `ensureTraderService` extraction, and `setupTraderAgent` consolidation into `tests/test-helpers.ts`.

Approving.

rajat2502 (approved):
LGTM. Lifts the polystrat-only `agentId == 86` gate so every Olas service on Polygon is indexed, with cohort filtering moved to deduplicated `TraderService.agentIds`/`operators` arrays (mirroring `pearl-transactions`), plus a `PayoutSource` enum on `PayoutRedemption`.

Reviewed: handler logic, dedup semantics, and reverse-link idempotency are correct; the sole `new TraderAgent` site and all test sites set the now-required `traderService`; all three `processRedemption` callers pass the new required `source`. No security-relevant changes.

Non-blocking items to keep in mind at rollout (already in the PR test plan):
- Schema change is breaking (TraderService immutable→mutable + new required fields, required `TraderAgent.traderService`) — needs a fresh full re-sync on a parallel deployment, not a graft.
- `Global.totalTraderAgents` semantics shift to cross-cohort; confirm consumers expect this or switch to `traderService_: { agentIds_contains: [86] }`.
- Indexing volume grows now that every Polygon Olas service becomes a TraderAgent — sanity-check before cutover.

Nit: `ensureTraderService` in `profit.test.ts` is declared between two import statements; consider relocating/sharing it.

PR #144, opened 2026-06-05
fix(pearl-transactions): hide registry dust + carry agent link on bond rows

rajat2502:
## Why

Two wallet-history correctness bugs surfaced during Gnosis QA of VLOP-73:

1. **Registry dust shown as deposits.** The `ServiceRegistryL2` contract sends 1-wei native dust to the Master Safe during terminate/unbond. `classifyTransfer` fell through to `MASTER_FUNDING_IN`, so the wallet rendered these as `+0.00 XDAI` **Deposit** rows (sharing a tx with the real `SERVICE_BOND_REFUND`).
2. **Bonds showed a generic "Agent" name** (e.g. "Agent stake/unstake" instead of "Omenstrat stake/unstake"). The wallet filters `FundsMovement` by `masterSafe`, so it saw the **RAW_TRANSFER** bond row (carries `masterSafe`, but no `service`/`agentSafe`) rather than the **SEMANTIC** bond row (carries `service` + `bondType`, but `masterSafe` was null).

## What

- **`getServiceRegistryAddress(network)`** added to `constants.ts` (mirrors `networks.json`, all 4 chains).
- **`classifyTransfer`**:
  - Inbound to a Master Safe **from the registry contract → `OTHER`** (kept for the forensic view, hidden from the wallet) instead of `MASTER_FUNDING_IN`.
  - Master Safe ↔ SRTU OLAS hops now **return `null`** — the RAW duplicate is dropped so the SEMANTIC row is the single bond record (no double-count in a `masterSafe`-filtered query). `TokenBalance` is unaffected (bonds were never in its update branches).
- **Bond rows are now self-contained.** The SRTU producer stamps `masterSafe` from the bond payer (`account`, guarded to tracked Master Safes); `dequeueAndAttribute` backfills `agentSafe` (+ `masterSafe` fallback) from the now-known `Service`. So one SEMANTIC row carries `masterSafe + service + bondType + agentSafe`. `agentSafe` resolves on **unstake** (multisig exists); a genesis **stake** leaves it null but `service.agentIds` still yields the display name.

## Tests

- The 2 "RAW reconciliation row" tests become **suppression** assertions (the hop no longer emits a RAW row).
- Adds a **registry-dust → OTHER** test and a **bond row carries masterSafe + agentSafe** test.
- `yarn test` → **46 pass**; `yarn build` clean.

## Consumer note

A complementary frontend change (read the agent name from the bond row's direct `service` when `agentSafe` is null, for the genesis-stake case) lands on the Operate-app VLOP-73 branch. Requires a **v0.0.6** redeploy to surface on-chain.

### DIFF
--- a/subgraphs/pearl-transactions/CLAUDE.md
+++ b/subgraphs/pearl-transactions/CLAUDE.md
@@ -49,7 +49,15 @@ deploy; watch Polygon USDC.e sync (the §2.2 cost hotspot).
   queue; dual-guarded Master Safe discovery on the NFT path.
 - **`ServiceRegistryTokenUtility`** (`src/service-registry-token-utility.ts`)
   — `TokenDeposit` / `TokenRefund`. Producer side: creates the
-  `SERVICE_BOND_DEPOSIT` / `_REFUND` row (amount only) + enqueues it.
+  `SERVICE_BOND_DEPOSIT` / `_REFUND` row + stamps `masterSafe` from the bond
+  payer (`account`, guarded to tracked Master Safes) + enqueues it. The
+  consumer (`dequeueAndAttribute`) backfills `serviceId` + `bondType` +
+  `agentSafe` (and `masterSafe` as fallback). This single SEMANTIC row is the
+  complete bond record — `classifyTransfer` drops the Master Safe ↔ SRTU OLAS
+  `RAW_TRANSFER` duplicate so the wallet (masterSafe-filtered) doesn't
+  double-count. `agentSafe` resolves on refunds (multisig already exists); a
+  genesis deposit leaves it null but `service.agentIds` still yields the
+  display name.
 - **`StakingFactory`** (`src/staking-factory.ts`) — `InstanceCreated`:
   allow-list check → spawn `StakingProxy` template + create `StakingContract`.
 - **`StakingProxy`** template (`src/staking-proxy.ts`) — `ServiceStaked`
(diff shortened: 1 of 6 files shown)


mohandast52 (approved):
### Suggestions

**1. Add a test for the `dequeueAndAttribute` masterSafe-fallback path**
`subgraphs/pearl-transactions/src/utils.ts` (line ~478)

The fallback only writes `movement.masterSafe` when `movement.masterSafe === null` — but the new bond-row test seeds the Master Safe before the refund, so `linkBondMasterSafe` always succeeds and the fallback branch stays dead in tests. A test that calls `handleTokenRefund` *before* the Master Safe is tracked would exercise that branch and lock it in against regressions if the producer guard ever shifts.

**2. Note the OLAS-vs-native test substitution**
`subgraphs/pearl-transactions/tests/phase-2a.test.ts` (line ~376)

The production bug is 1-wei native (xDAI) dust, but the test simulates it with `newOlasTransfer`. The classifier is token-agnostic so coverage is equivalent, but a one-line comment ("OLAS exercises the same classifier path as native dust") would close the gap between the test and the bug it's preventing.

### Questions

- `subgraphs/pearl-transactions/src/utils.ts` (line ~487): the backfill is asymmetric — `masterSafe` only fills when null, but `agentSafe` is unconditionally overwritten whenever `service.agentSafe != null`. I read this as intentional because `agentSafe` is only written here, but a one-line comment explaining the asymmetry would save future maintainers a trace through the code.

### Highlights

- Producer/consumer dual-stamping is the right shape for this bug — neither side alone has all four fields, so stamping locally with what's known + falling back in the consumer is correct.
- The `MASTER`-role guard in `linkBondMasterSafe` keeps non-Pearl bonds on the same SRTU contract unlinked — solid defense.
- Rewriting the existing tests as `assert.entityCount("FundsMovement", 0)` suppression assertions is clean — same coverage shape with the flipped expectation.
- The CLAUDE.md update explains both the producer stamping and the RAW dedup. Genuinely useful for future investigators.

kupermind (approved):

PR #145, opened 2026-06-08
fix(pearl-transactions): ServicesEvicted ID collision + token decimals fallback

rajat2502:
Addresses correctness findings raised by @DIvyaNautiyal07 in the review of the self-hosted Gnosis port ([valory-xyz/autonolas-subgraph#102](https://github.com/valory-xyz/autonolas-subgraph/pull/102)). Fixing here in the canonical source first per the sync policy; the port re-mirrors after this merges.

## P1 — correctness

- **`ServicesEvicted` row-ID collision** (`staking-proxy.ts`). The per-slot id was `txHash.concatI32(logIndex + i)`. A `ServicesEvicted` at logIndex `N` affecting `K` services claimed ids `N..N+K-1`, so any other event in the same tx at a logIndex in `[N+1, N+K-1]` produced the same id and **silently overwrote** its `FundsMovement` row. Now composes logIndex and the slot index as **separate** segments (`concatI32(logIndex).concatI32(i)`).
- **`getOrCreateToken` decimals fallback** (`utils.ts`). The no-resolver branch persisted `UNKNOWN/18`. This branch is only reachable for an *indexed* token, and every indexed non-OLAS/non-wrapped token is a **6-decimal stablecoin**, so an 18-dec fallback misformats amounts by 10^12 — and first-write-wins makes it permanent. We can't rely on `log.critical` aborting (graph-node may continue), so the persisted value must be safe by itself: **default to 6**. Kept the loud `log.critical` as the drift canary.

## Clear nits

- Collapse the `isDelta ? amount : amount` identical-branch no-op in `upsertTokenBalance`.
- Drop the unused `masterSafePtr` param from `classifyTransfer` (every caller passed `null`) + update the two callers.

## Tests

Eviction test updated to the new compound id. `yarn build` ✅, `yarn test` → **46 pass** ✅.

## Follow-ups (not in this PR)

The P2 hardening items from the review (allow-list silent skip, operator-EOA tracking gap, `drainPendingRegistration` cleanup, `Address.zero()` phantom refs, masterEoa-rotation test gap) are tracked for a separate batch. Native-out placeholder remains the documented v1 limitation.

> Note: ships in a future redeploy (v0.0.7) of the 4 Studio subgraphs + re-mirror into the self-hosted port.

### DIFF
--- a/subgraphs/pearl-transactions/src/utils.ts
+++ b/subgraphs/pearl-transactions/src/utils.ts
@@ -648,16 +648,19 @@ export function getOrCreateToken(tokenAddress: Address): Token {
       // Reached here only via a tracked Transfer, so this is an indexed
       // token with no metadata resolver — almost certainly an
       // `erc20Tokens` entry in networks.json without a matching
-      // getStablecoinSymbol branch. UNKNOWN/18 would silently misformat a
-      // 6-decimal stablecoin by 10^12 in every consumer, and first-write-
-      // wins makes it permanent. Loud on purpose so it surfaces in
-      // indexing logs, not in the wallet UI.
+      // getStablecoinSymbol branch. Every such indexed token (OLAS and
+      // wrapped-native are handled above) is a stablecoin, so default to
+      // 6 decimals — NOT 18: an 18-decimal fallback would silently
+      // misformat the amount by 10^12 in every consumer, and first-write-
+      // wins makes it permanent. log.critical is loud on purpose (drift
+      // canary), but we must not rely on it aborting — graph-node may
+      // continue — so the persisted value has to be safe by itself.
       log.critical(
-        "getOrCreateToken: no symbol/decimals resolver for indexed token {} on {} — add a getStablecoinSymbol branch (networks.json `erc20Tokens` and the resolver are out of sync). Falling back to UNKNOWN/18.",
+        "getOrCreateToken: no symbol/decimals resolver for indexed token {} on {} — add a getStablecoinSymbol branch (networks.json `erc20Tokens` and the resolver are out of sync). Defaulting to UNKNOWN/6.",
         [tokenAddress.toHexString(), network]
       );
       token.symbol = "UNKNOWN";
-      token.decimals = 18;
+      token.decimals = 6;
     }
   }
   token.save();
@@ -680,7 +683,9 @@ export function upsertTokenBalance(
     bal = new TokenBalance(id);
     bal.safe = safe;
     bal.token = getOrCreateToken(tokenAddress).id;
-    bal.balance = isDelta ? amount : amount;
+    // First sighting: the initial balance is `amount` whether it's the
+    // first signed delta or an absolute baseline write.
+    bal.balance = amount;
   } else {
     bal.balance = isDelta ? bal.balance.plus(amount) : amount;
   }
@@ -721,8 +726,7 @@ export class ClassifyResult {
 // silently dropped").
 export function classifyTransfer(
   from: Address,
-  to: Address,
-  masterSafePtr: MasterSafe | null
+  to: Address
 ): ClassifyResult | null {
   const fromMaster = TrackedSafe.load(from);
   const toMaster = TrackedSafe.load(to);
--- a/subgraphs/pearl-transactions/CLAUDE.md
+++ b/subgraphs/pearl-transactions/CLAUDE.md
@@ -102,7 +102,7 @@ deploy; watch Polygon USDC.e sync (the §2.2 cost hotspot).
   in constants.ts).
 - **`getOrCreateToken`** — OLAS / wrapped-native 18 decimals; stablecoins
   (USDC / USDC.e / pUSD) 6 decimals via `getStablecoinSymbol` (constants.ts);
-  `log.critical` + UNKNOWN/18 if an indexed token has no resolver branch.
+  `log.critical` + UNKNOWN/6 if an indexed token has no resolver branch.
 - **`AgentFundingEvent`** — groups same-tx `MASTER_TO_AGENT` rows so one
   funding action is one consumer row.
 - **`SAFE_SETUP_TRANSFER` / Path A** — the first live Master-EOA → Master-Safe
--- a/subgraphs/pearl-transactions/src/erc20.ts
+++ b/subgraphs/pearl-transactions/src/erc20.ts
@@ -32,7 +32,7 @@ export function handleErc20Transfer(event: TransferEvent): void {
   const amount = event.params.value;
   const token = event.address;
 
-  const classification = classifyTransfer(from, to, null);
+  const classification = classifyTransfer(from, to);
   if (classification === null) {
     return;
   }
--- a/subgraphs/pearl-transactions/src/safe.ts
+++ b/subgraphs/pearl-transactions/src/safe.ts
@@ -30,7 +30,7 @@ export function handleSafeReceived(event: SafeReceivedEvent): void {
   const to = event.address;
   const amount = event.params.value;
 
-  const classification = classifyTransfer(from, to, null);
+  const classification = classifyTransfer(from, to);
   if (classification === null) return;
 
   const row = new FundsMovement(fundsMovementId(event));
--- a/subgraphs/pearl-transactions/src/staking-proxy.ts
+++ b/subgraphs/pearl-transactions/src/staking-proxy.ts
@@ -185,8 +185,13 @@ export function handleServicesEvicted(event: ServicesEvictedEvent): void {
     const service = getOrCreateService(serviceId, event);
 
     // Unique-id per (tx, logIndex, serviceId-slot) since ServicesEvicted
-    // is one event affecting many services. Use logIndex + i.
-    const id = event.transaction.hash.concatI32(event.logIndex.toI32() + i);
+    // is one event affecting many services. Compose logIndex AND the slot
+    // index as separate segments — `logIndex + i` would collide with any
+    // other event in the same tx whose logIndex falls in [logIndex+1,
+    // logIndex+K-1] and silently overwrite its FundsMovement row.
+    const id = event.transaction.hash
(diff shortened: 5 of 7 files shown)


kupermind:
## Review — analysis

Reviewed the diff against `main`. Both P1 fixes are correct and the nits are clean. **LGTM.**

### P1.1 — `ServicesEvicted` row-ID collision ✅
Verified the collision is real, not theoretical. `fundsMovementId(event)` (`utils.ts:48`) is `txHash.concatI32(logIndex)` and is the id used by **every other** `FundsMovement` producer — `erc20.ts:40`, `safe.ts:36`, the two staking-reward rows (`staking-proxy.ts:71,111`), and both SRTU bond rows (`service-registry-token-utility.ts:36,56`). The old eviction id `txHash.concatI32(logIndex + i)` therefore aliases the single-segment id of any same-tx event sitting at `logIndex+i` for `i ≥ 1`, silently overwriting it (first-write loses).

The fix `concatI32(logIndex).concatI32(i)` is sound on two axes:
- vs. single-segment ids: the result is 4 bytes longer, so it can never equal a `txHash.concatI32(x)` id regardless of value.
- vs. another eviction in the same tx: the middle `logIndex` segment differs, so no cross-eviction collision.

I checked that no other code path builds a two-segment `txHash.concatI32(_).concatI32(_)` id (`utils.ts:76` is single-segment), so the new shape is collision-free. Test update (`tx.concatI32(0).concatI32(0/1)`) matches the new construction for the mock's logIndex 0.

### P1.2 — `getOrCreateToken` decimals fallback ✅
Agree with defaulting the no-resolver branch to **6** rather than 18, and with keeping `log.critical` as the drift canary (can't rely on it aborting — graph-node may continue, and first-write-wins makes the persisted value permanent). For the current token set (OLAS + wrapped-native handled above; every remaining indexed token is a 6-dec stablecoin) this is strictly safer.

One thing worth keeping in mind (not blocking): this hard-codes the assumption "any future drift token is also 6-dec." A DAI-class 18-decimal stablecoin added to `erc20Tokens` without a `getStablecoinSymbol` branch would now be the misformat case instead. The canary still fires loudly so it surfaces, and the lockstep `networks.json` + `getStablecoinSymbol` rule in the CLAUDE.md is the real guard — just noting the residual edge so the canary isn't ignored.

### Nits ✅
- `bal.balance = isDelta ? amount : amount` → `bal.balance = amount`: identical-branch no-op, correct on first-sighting (initial balance is `amount` either way).
- Dropping `masterSafePtr`: confirmed it appears only in the signature, never in the body; both callers passed `null`; `MasterSafe` import stays live (used at `utils.ts:100/125/183/822/894`), so no orphaned import.

### Meta
- CI: all checks green (test/pearl-transactions + the full matrix).
- Eviction rows get a new id shape, so deployed v0.0.5 rows re-key on resync — consistent with the noted v0.0.7 redeploy + re-mirror. No consumer reads these ids by value, so no downstream impact.

Nothing to change. The deferred P2 items (allow-list silent skip, operator-EOA gap, `drainPendingRegistration`, `Address.zero()` phantom refs, masterEoa-rotation test) and the native-out v1 limitation are correctly scoped out.

rajat2502:
Thanks @kupermind — appreciate the thorough verification, especially confirming the eviction-ID alias against every other single-segment `FundsMovement` producer.

On the non-blocking P1.2 note: agreed, and good to flag the residual edge explicitly. The defence-in-depth stays the same — the loud `log.critical` canary fires on any unresolved indexed token, and the lockstep `networks.json` `erc20Tokens` + `getStablecoinSymbol` rule (CLAUDE.md "Honest limits") is the real guard against drift. Defaulting to 6 just makes the *common* drift case (a new 6-dec stablecoin) safe-by-default instead of silently misformatted; an 18-dec addition would still trip the canary loudly before anyone trusts the value. Leaving as-is.

No changes — will let this ride into the v0.0.7 redeploy + re-mirror to the self-hosted port.

DIvyaNautiyal07 (commented):
**P1 — Stale `UNKNOWN/18` references (ripple of the decimals fix)**

The PR changes the fallback from `UNKNOWN/18` to `UNKNOWN/6`, but two references to the old value were not updated:

- `subgraphs/pearl-transactions/CLAUDE.md:105` — "`log.critical` + UNKNOWN/18 if an indexed token has no resolver branch." Should be `UNKNOWN/6`.
- `subgraphs/pearl-transactions/tests/phase-2a.test.ts:760` — comment says "instead of silently mapping to UNKNOWN/18 in production." Should be `UNKNOWN/6`.

---

**P2 — Eviction test doesn't actually exercise the collision the fix prevents**

`tests/staking.test.ts:577-588` updates the expected ID format, so reverting the fix would make the test fail. But the test fires a single `ServicesEvicted` at mock `logIndex=0` and never fires a second event in the same tx at the would-collide `logIndex`. The exact corruption scenario the PR fixes (a same-tx `FundsMovement` at `logIndex+i` getting silently overwritten by the eviction's slot-i row under the old `concatI32(logIndex + i)` scheme) is not covered.

Suggest adding a test that fires an eviction at `logIndex=2` with 2+ services AND another `FundsMovement`-producing event at `logIndex=3` in the same tx, then asserts `entityCount("FundsMovement", N)` and that both event types' rows exist.

---

**P2 — No test exercises the `else` fallback branch in `getOrCreateToken`**

The branch where `decimals = 6` is set (utils.ts:663) is unreachable today — every `erc20Tokens` entry is matched by `getStablecoinSymbol`. If a regression flipped the fallback back to 18, no existing test would fail. Low real impact, but worth a unit test passing a synthetic address that misses all three branches and asserting `decimals = "6"`.

rajat2502:
Thanks @DIvyaNautiyal07 — addressed in `fc20bb3`:

- **P1 stale `UNKNOWN/18` refs** — fixed both to `UNKNOWN/6` (`CLAUDE.md:105` and the Phase-2b test comment).
- **P2 collision regression test** — added. Fires a 2-service `ServicesEvicted` at `logIndex 2` + a same-tx `RewardClaimed` at `logIndex 3`, then asserts `entityCount("FundsMovement", 3)` and that both event types' rows survive. It genuinely guards the fix: under the old `concatI32(logIndex + i)` scheme the eviction's slot-1 row (`2+1=3`) aliased the reward's single-segment id `concatI32(3)` → one overwrote the other → only 2 rows. Confirmed it fails if I revert the production change.

- **P2 `getOrCreateToken` fallback unit test** — I tried it (synthetic non-resolver address → assert `decimals = 6`), but it's **not implementable**: `log.critical` is fatal in matchstick — it panics before the `Token` is saved, so the assert is unreachable:
  ```
  thread 'main' panicked at '🆘 getOrCreateToken: no symbol/decimals resolver … Defaulting to UNKNOWN/6.'
  ```
  A `shouldFail` test would only prove the canary fires, not that the persisted value is `6` vs `18` (both would panic identically), so it wouldn't catch the regression you're worried about. The only way to make it assertable is downgrading the canary to `log.error`, which weakens the production drift signal — not worth it. Leaving the branch covered by the code + the lockstep `networks.json`/`getStablecoinSymbol` rule. Reverted that test.

47 tests pass.

DIvyaNautiyal07 (approved):

subgraphs/pearl-transactions/src/utils.ts
```diff
@@ -648,16 +648,19 @@ export function getOrCreateToken(tokenAddress: Address): Token {
...
-        "getOrCreateToken: no symbol/decimals resolver for indexed token {} on {} — add a getStablecoinSymbol branch (networks.json `erc20Tokens` and the resolver are out of sync). Falling back to UNKNOWN/18.",
+        "getOrCreateToken: no symbol/decimals resolver for indexed token {} on {} — add a getStablecoinSymbol branch (networks.json `erc20Tokens` and the resolver are out of sync). Defaulting to UNKNOWN/6.",
         [tokenAddress.toHexString(), network]
       );
       token.symbol = "UNKNOWN";
-      token.decimals = 18;
+      token.decimals = 6;
     }
   }
   token.save();
```
DIvyaNautiyal07, line 666: Good-to-have: one stale `UNKNOWN/18` reference outside this PR's diff — `scripts/generate-manifests.js:75` says "fall back to UNKNOWN/18" in the same comment style as the CLAUDE.md ripple you already did. Worth flipping to `UNKNOWN/6` so the generator's docs stay in sync with the fallback.

DIvyaNautiyal07 (approved):

Tanya-atatakai (approved):
Reviewed both P1 fixes in context and verified locally: `yarn build` ✅, `yarn test` → all **47 pass** ✅.

**ServicesEvicted ID collision** — fix is sound on two levels: the two-segment id is 40 bytes vs. 36 for every other `FundsMovement` id (so cross-handler collision is structurally impossible), and `(txHash, logIndex, i)` is unique by construction within eviction rows. Confirmed the new regression test reproduces the exact old-scheme collision (eviction at logIndex 2, `RewardClaimed` at logIndex 3 → byte-identical ids) and would have caught the bug. One consequence to keep in mind (not a blocker): existing eviction rows get new ids, which is fine for the from-scratch v0.0.7 resync but would break a graft if grafting ever enters the deploy flow.

**Decimals fallback 18 → 6** — reasoning holds: the branch is only reachable for an allowlisted `erc20Tokens` token, and every such non-OLAS/non-wrapped token is a 6-dec stablecoin. The residual risk is symmetric (a future 18-dec allowlist addition without a resolver branch misformats the other way), but the Phase-2b suite covering all 8 (chain, token) tuples + the `log.critical` canary is the right defense.

**Cleanups** — `isDelta ? amount : amount` collapse is a pure no-op removal; `masterSafePtr` drop checked against all call sites (only `erc20.ts` + `safe.ts`, both updated, no direct test callers).

**Nit**: the `CLAUDE.md` Tests section still says "46 Matchstick tests" — the new regression test makes it 47 (the PR body's "46 pass" likely predates that commit). One-line bump per the repo's CLAUDE.md-maintenance convention.

PR #147, opened 2026-06-10
feat(pearl-transactions): split OLAS agent→master into AGENT_OLAS_TO_MASTER

rajat2502:
## What

Adds a new `FundsCategory` — **`AGENT_OLAS_TO_MASTER`** — for OLAS Agent Safe → Master Safe transfers, splitting them out of `AGENT_TO_MASTER`.

These hops are dominated by **staking-reward sweeps** (the agent Safe receives a `RewardClaimed` payout, then a separate tx sweeps the same OLAS amount up to the Master Safe ~10–15s later), plus any manual OLAS return. They aren't user wallet actions, and the sweep tx emits only a bare `Transfer` + `ExecutionSuccess` — there is **no on-chain marker** to tell a reward sweep from a genuine manual OLAS return. So we bucket them together by token rather than guessing.

## Why

The Pearl wallet history wants to hide the reward-sweep noise. Previously the frontend fetched `AGENT_TO_MASTER` and filtered OLAS rows client-side. Giving OLAS agent→master its own category lets the wallet **exclude it at query time** (`category_in` without `AGENT_OLAS_TO_MASTER`) instead of fetch-then-filter — less data over the wire, simpler consumer.

Native / non-OLAS agent→master is genuine and **stays `AGENT_TO_MASTER`**.

## How

- `schema.graphql` — new `AGENT_OLAS_TO_MASTER` enum value.
- `constants.ts` — `CATEGORY_AGENT_OLAS_TO_MASTER`.
- `erc20.ts` — after `classifyTransfer`, re-tag an `AGENT_TO_MASTER` hop to `AGENT_OLAS_TO_MASTER` when `token == OLAS`. The token isn't visible inside `classifyTransfer` (it only sees from/to), so the split happens in the handler. `TokenBalance` accounting is unchanged — the new category appears in the same debit/credit branches as `AGENT_TO_MASTER`.

## Tests

- Updated the existing OLAS agent→master test to assert `AGENT_OLAS_TO_MASTER`.
- Added a non-OLAS (USDC) agent→master test proving the split is token-gated (stays `AGENT_TO_MASTER`).
- All 47 Matchstick tests pass; `codegen` + `build` green.

## Follow-ups (not in this PR)

- Frontend: drop the client-side OLAS filter; omit `AGENT_OLAS_TO_MASTER` from `category_in`.
- Redeploy (v0.0.7) + re-mirror to the self-hosted Gnosis port.


kupermind:
## Review — analysis

Reviewed the diff against `main`. The split is correct, balance-neutral, and well-tested. **LGTM.**

### Correctness ✅
- **Re-tag placement.** `classifyTransfer` only sees `(from, to)`, so it can't know the token — doing the OLAS split in the handler after classification is the right seam. The re-tag is gated on `category == AGENT_TO_MASTER && token == OLAS`, so only the OLAS leg moves; native / stablecoin agent→master stays `AGENT_TO_MASTER`. Verified against the gnosis OLAS address in `getOlasAddress(currentNetwork())`.
- **No missed consumers.** Grepped every `AGENT_TO_MASTER` reference: `utils.ts:789` is the classification *source* (just assigns the string), and the only *consumers* of the category value are the two `TokenBalance` branches in `erc20.ts` — both of which the PR extends with `AGENT_OLAS_TO_MASTER`. Nothing else (no `DailyServiceFunds` aggregation, no `AgentFundingEvent` path, no `safe.ts` branch) keys off this category, so re-tagged rows can't silently fall through.
- **Balance accounting is genuinely unchanged.** For an OLAS agent→master both sides are tracked: the credit branch bumps the Master Safe (`to`, +amount) and the debit branch bumps the Agent Safe (`from`, -amount). `AGENT_OLAS_TO_MASTER` is added to *both* branches, so it mirrors `AGENT_TO_MASTER` exactly — only the label differs, as the comment claims.

### Tests ✅
The token-gating is the load-bearing claim and it's now covered both ways: OLAS → `AGENT_OLAS_TO_MASTER`, USDC (`0xDDAfbb50…`, a tracked non-OLAS token) → stays `AGENT_TO_MASTER`. CI green (test/pearl-transactions + full matrix, 47 tests).

### Notes (non-blocking)
- **Stale comment at `utils.ts:781`** (not touched by this PR): the classifier still annotates the agent→master branch as `// → AGENT_TO_MASTER (OLAS reward sweep)`. Post-merge the OLAS sweep is no longer what lands in `AGENT_TO_MASTER` — worth a one-line follow-up so the source-of-truth comment doesn't mislead.
- **Deploy ordering.** This bucket *hides* OLAS agent→master (sweeps **and** manual returns — the PR explicitly accepts not distinguishing them). After v0.0.7 lands but before the frontend follow-up, OLAS rows simply vanish from the existing `AGENT_TO_MASTER` query rather than being client-side-filtered — which is the intended "hide the noise" behavior, so the interim state is safe. Existing v0.0.5 rows keep the old label until resync. Flagging only so the frontend change isn't treated as strictly-after-optional.
- **Stacks cleanly on #145** — that PR edits `erc20.ts:35` (the `classifyTransfer` arity) and `utils.ts`; this one edits the imports + the post-classification block + the balance branches. Disjoint lines, clean auto-merge either order.

Nothing to change here.

DIvyaNautiyal07 (commented):
**P1 — Should fix**

- `subgraphs/pearl-transactions/README.md:30` — The `FundsCategory` table still reads `AGENT_TO_MASTER | Agent Safe → Master Safe (reward sweep)`. After this PR that description is wrong (the reward sweep is `AGENT_OLAS_TO_MASTER`), and the new category is missing from the table entirely. This README is the externally-consumed quick reference. Suggest updating the existing row to "native / non-OLAS only" and adding a new row for `AGENT_OLAS_TO_MASTER`.

- `subgraphs/pearl-transactions/tests/phase-2a.test.ts:423-468` — Neither new test asserts the `TokenBalance` side-effect for `AGENT_OLAS_TO_MASTER`. `erc20.ts:109-135` adds the new category to both the credit and debit branches, but if either insertion were dropped both new tests would still pass — they only check `FundsMovement.category`. Worth an `assert.fieldEquals("TokenBalance", ...)` on the agent and master sides in the positive test (the pattern already exists elsewhere in the file).

**P2 — Consider**

- `subgraphs/pearl-transactions/src/utils.ts:781` — `classifyTransfer`'s inline comment still says `Agent Safe → Master Safe → AGENT_TO_MASTER (OLAS reward sweep)`. After the split, OLAS reward sweeps don't end up in `AGENT_TO_MASTER`; they're re-tagged in `erc20.ts`. Either drop the qualifier or note the re-tag.

- `subgraphs/pearl-transactions/src/erc20.ts:140` — Trailing comment `// AGENT_TO_MASTER also updates Master side; handled above for "to".` doesn't mention `AGENT_OLAS_TO_MASTER`, which is now in the same family. Mention both, or drop the comment.

- `subgraphs/pearl-transactions/tests/phase-2a.test.ts:432-438` — The positive test asserts only `category == "AGENT_OLAS_TO_MASTER"`. It doesn't assert `row.token` equals the gnosis OLAS address. If `newOlasTransfer` were ever changed to emit at a non-OLAS address, the category assertion could still pass for an unrelated reason. Adding a `token` field assertion would lock the contract.

**Nit**

- `subgraphs/pearl-transactions/schema.graphql:97` — The new enum line carries a long-form rationale ("we do NOT distinguish reward sweep vs manual return") that wraps visually compared to neighboring one-line labels. The rationale is already in `CLAUDE.md` and the `erc20.ts` comment, so a short label here would match the existing style better.

rajat2502:
Thanks @DIvyaNautiyal07 — all addressed in `2f3bfcf`:

**P1**
- `README.md` FundsCategory table — `AGENT_TO_MASTER` updated to "native / non-OLAS token" and a new `AGENT_OLAS_TO_MASTER` row added.
- `tests/phase-2a.test.ts` — the OLAS positive test now asserts the `TokenBalance` side-effect on **both** sides (Master credited `+AMOUNT`, Agent debited `-AMOUNT`). Confirmed it's load-bearing: dropping either branch in `erc20.ts:109-135` now fails the test.

**P2**
- Same test now also asserts `row.token == OLAS_GNOSIS`, so the category can't pass for an unrelated address.
- `utils.ts` classifier comment refreshed — notes the OLAS leg is re-tagged in `handleErc20Transfer` and native/non-OLAS stays `AGENT_TO_MASTER`.
- `erc20.ts` trailing `TokenBalance` comment now mentions `AGENT_OLAS_TO_MASTER`.

**Nit**
- Shortened the `schema.graphql` enum comment to a one-liner matching its neighbors (full rationale already in `CLAUDE.md` + `erc20.ts`).

47 tests pass; build green.

Also flagging @kupermind's non-blocking note (stale `utils.ts` comment) — covered by the same P2 comment refresh.

subgraphs/pearl-transactions/schema.graphql
```diff
@@ -93,7 +93,8 @@ enum FundsCategory {
   MASTER_FUNDING_IN                   # subsequent EOA → Master Safe (after SAFE_SETUP_TRANSFER consumed)
   MASTER_WITHDRAWAL                   # Master Safe → EOA
   MASTER_TO_AGENT                     # Master Safe → Agent Safe / Agent EOA (grouped via AgentFundingEvent)
-  AGENT_TO_MASTER                     # Agent Safe → Master Safe (OLAS reward sweep)
+  AGENT_TO_MASTER                     # Agent Safe → Master Safe in native / non-OLAS token
+  AGENT_OLAS_TO_MASTER                # Agent Safe → Master Safe in OLAS (reward sweeps + manual returns); split out so the wallet can exclude it at query time
```
DIvyaNautiyal07, line 97: Good-to-have: a few cross-subgraph ripple references to `AGENT_TO_MASTER` in `subgraphs/pearl-funds/IMPLEMENTATION-PLAN.md` are now stale for the OLAS-sweep case — specifically lines 454 ("optional OLAS sweep → AGENT_TO_MASTER (Phase 2a raw)"), 505 ("optional profit sweep<br/>AGENT_TO_MASTER"), and 828 ("Agent Safe → Master Safe ⇒ AGENT_TO_MASTER (the reward sweep)"). Internal planning doc only, not consumer-facing, so safe as a follow-up — flagging so it's on the radar with the v0.0.7 redeploy.

DIvyaNautiyal07 (approved):

Tanya-atatakai (approved):
## Review — approve ✅

### Correctness (verified locally: codegen + build + all 47 Matchstick tests pass)

- The re-tag in `erc20.ts` is placed correctly: right after the classification null-check and before the row is built, so `row.category` and both `TokenBalance` branches see the final category. The `SAFE_SETUP_TRANSFER` / `MASTER_TO_AGENT` side-effect blocks can never match an agent→master hop, so they're unaffected.
- Token gate is sound: `getOlasAddress` covers all 4 deployed networks (incl. `xdai`/`polygon` aliases), and the `getOlasAddress(currentNetwork())` pattern is already proven in `getOrCreateToken`.
- No missed call sites: the only other `classifyTransfer` consumer is `safe.ts` (native `SafeReceived`), which by design keeps native agent→master as `AGENT_TO_MASTER` and has no `TokenBalance` branches. Nothing else in `src/` keys off `AGENT_TO_MASTER`.
- Balance accounting preserved: `AGENT_OLAS_TO_MASTER` mirrors `AGENT_TO_MASTER` in both the credit-`to` and debit-`from` branches, and the updated test locks in both `TokenBalance` sides.
- Test pair is right: positive test asserts category + token + both balances; the USDC negative test proves the split is token-gated, not route-gated.
- Deploy sequencing is benign: until the frontend updates `category_in`, its existing client-side OLAS filter just becomes a no-op — no window where sweeps wrongly reappear.

### Before merging

1. **CI is not fully green** — `test / governance` failed with `Failed to get matchstick binary: Invalid version. Must be a string.` This is an infra issue in the governance job's Matchstick binary fetch, unrelated to this PR (`test / pearl-transactions` passed). Please re-run the failed job so the merge goes in on a green run.
2. **Branch is behind `main`** — missing the 4 commits from #146 (license allowlist gate, incl. the `d89fd3d` SPDX fix). Please merge `main` in (or rebase) so the new License-gate CI runs against this branch's tree before merge.

### Nits (non-blocking)

- `subgraphs/pearl-transactions/CLAUDE.md` "## Tests" section still says **46** Matchstick tests — this PR makes it 47.
- The re-tag uses the `CATEGORY_*` constants while the `TokenBalance` branches below use string literals (pre-existing style); a follow-up converging the file on the constants would remove the drift risk.

PR #148, opened 2026-06-11
docs(pearl-transactions): indexing performance review

Tanya-atatakai:
## Summary

Adds `subgraphs/pearl-transactions/INDEXING-PERFORMANCE.md` — a review of the subgraph's indexing speed against The Graph's best practices (pruning, `@derivedFrom`, immutable entities + Bytes IDs, eth_calls, timeseries, grafting). Documentation only; no schema, mapping, or manifest changes.

## Contents

- **TL;DR**: sync time is dominated by the chain-wide ERC-20 `Transfer` data sources — `classifyTransfer` does up to 6 store lookups per transfer to discard ~99.99% of events.
- **Ranked proposals**:
  1. Raise OLAS / wrapped-native / stablecoin start blocks to the first Pearl Safe per chain (zero code, graftable; includes the `prune: auto` × graft-block caveat).
  2. Merge `TrackedSafe` / `TrackedEOA` / staking lookup into one `TrackedAddress` entity — 6 loads → 2 on the hot path.
  3. Mark write-once entities (`TrackedSafe`/`TrackedEOA`/`StakingContract`/`Token`) `immutable: true`.
  4. Split mutable bond rows out of `FundsMovement` so the high-cardinality ledger can go immutable (consumer-facing; needs wallet-team sign-off), plus the `Service` Bytes-ID gap.
  5. Spec bump notes — including why declared eth_calls would be a regression here.
- **Checked and fine**: pruning, arrays/`@derivedFrom`, eth_call usage, and a considered-and-rejected timeseries conversion of `DailyServiceFunds`.
- **Polygon context**: why trigger density on USDC-class contracts makes Polygon the slowest chain to sync, and the self-hosted (Firehose/RPC/tuning) levers.
- **Suggested sequencing**: graftable `networks.json` change first, single clean-sync schema release second.

## Test plan

Docs-only change — no build/test impact. Reviewed against the live `schema.graphql`, `subgraph.template.yaml`, `networks.json`, and all `src/` mappings at v0.0.5.


rajat2502 (approved):

subgraphs/pearl-transactions/INDEXING-PERFORMANCE.md
```diff
@@ -0,0 +1,197 @@
...
+concluding "not ours", and ~99.99% of them are not ours. The two
+highest-leverage fixes:
+
+1. Raise the token data-source start blocks to the first Pearl Safe per
+   chain (zero code).
+2. Collapse the tracked-address lookup from 6 loads to 2, against
+   immutable tables.
+
+## Proposals, ranked by impact
+
```
DIvyaNautiyal07, line 23: Good-to-have: this justification has a small gap. `classifyTransfer` also returns `CATEGORY_OTHER` when either side equals the SRTU address (constant) or hits a `StakingContract` row, and `StakingContract` entities are created from `StakingFactory.InstanceCreated` independently of any Pearl Safe (see the `fromIsTracked || toIsTracked` fallback in `src/utils.ts`). So pre-MasterSafe transfers touching SRTU or a staking proxy *can* classify, just as `OTHER`. The proposal is still safe because those `OTHER` rows are non-Pearl noise we want to skip anyway, but the tighter framing is "no *Pearl-relevant* transfer can classify before the first MasterSafe."

DIvyaNautiyal07 (approved):

kupermind:
## Review — analysis

Docs-only, but a perf review is only as good as its claims, so I fact-checked each one against the v0.0.5 code rather than the prose. **It holds up very well — every factual assertion is accurate.** One substantive caveat on Proposal #1's chosen bound, plus a couple of minor notes; none block merge.

### Accuracy audit (verified against code)

| Claim | Verdict |
|---|---|
| `classifyTransfer` does 6 store loads + 2 address-string parses on the discard path | ✅ `utils.ts:727-736` — `TrackedSafe`×2, `TrackedEOA`×2, `StakingContract`×2, plus unconditional `getSrtuAddress()` + `getServiceRegistryAddress()` |
| `getServiceRegistryAddress()` computed at top but used in one branch | ✅ `registryAddr` defined `:736`, used only at `:819` (registry-dust check) |
| Token sources all start at the `ServiceRegistryL2` block | ✅ `networks.json` — OLAS/WrappedNative/erc20Tokens == registry block on every chain |
| `TrackedSafe`/`TrackedEOA`/`StakingContract`/`Token` write-once but `immutable: false` | ✅ schema `:223,232,242,68`; `upsertTrackedEOA` early-returns on existing |
| `FundsMovement` mutable solely for SRTU bond backfill | ✅ schema `:119-123` + `dequeueAndAttribute` |
| `Service.id` / `DailyServiceFunds.id` are strings; `FundsMovement.id` / `ServiceIndex.id` Bytes; `ServiceIndex` uses `Bytes.fromBigInt` | ✅ schema `:50,148,124,172` + `utils.ts:61` |
| `getOwners()`/`getThreshold()` fire once per first sighting behind guards | ✅ `try_getOwners`/`try_getThreshold` only in the create branch after `MasterSafe.load` null-check |
| `getOrCreateMasterSafe` load+save on every re-sighting to bump `lastActivityTimestamp` | ✅ `utils.ts:100-104` |
| specVersion `1.0.0` / apiVersion `0.0.7`; `prune: auto` set | ✅ `subgraph.template.yaml:1,3,16…` |
| Double-saves (`getOrCreatePendingBondCounter`→`enqueuePendingBondRow`; `handleCreateMultisigWithAgents` ×2; `handleServiceStaked` ×2) | ✅ `utils.ts:424/445`, `service-registry.ts:93/97`, `staking-proxy.ts:42/55` |
| `@aggregate` can't express the `cumulativeOlasRewardsClaimed` running total | ✅ correct — interval funcs are sum/count/min/max/first/last/avg only |

### Substantive caveat — Proposal #1's bound isn't airtight as stated

The proposal's invariant is right ("no transfer can classify before the first tracked address exists"), but it operationalizes it as `MIN(MasterSafe.firstSeenBlock)`, and those two aren't provably equal. `classifyTransfer`'s fallback (`utils.ts:868-885`) returns a non-null `OTHER` row whenever **any** side is tracked — and two of those tracked predicates don't depend on a `MasterSafe`:

- **`StakingContract`** rows are created independently in `handleInstanceCreated` (`StakingFactory.InstanceCreated`), and the factory start block is kept as-is. A staking proxy can be deployed/funded (e.g. DAO → proxy OLAS) **before** the first Pearl Master Safe onboards on that chain; such a transfer hits the `toIsStaking` fallback → `OTHER` row.
- **SRTU / registry addresses** are constant-address equality checks (`fromIsSrtu`/`registryAddr`), "tracked" from block 0 with no entity at all.

So the strictly-safe lower bound is `MIN(first TrackedSafe block, first TrackedEOA block, MIN(StakingContract.createdBlock))` — not `MasterSafe.firstSeenBlock` alone. In practice the Master Safe is almost certainly earliest (and the dropped rows would be forensic-only `OTHER` with `masterRef = null`, which the wallet hides), so the proposal's likely fine — but I'd either **take the MIN across master-safe + staking-contract creation blocks**, or have the doc explicitly state it's accepting the drop of any pre-onboarding staking/SRTU `OTHER` rows. As written, "equally provable" slightly overstates it.

### Minor

- "Cuts per-trigger work ~3×" (Proposal #2) is the load-count ratio (6→2); actual wall-clock gain is less since loads aren't 100% of per-trigger work. Fine as an order-of-magnitude framing, just not literally 3× end-to-end.
- The declared-eth_calls warning in #5 is the sharpest call in the doc and it's correct — declared calls run before *every* matching handler invocation, which would convert the current once-per-first-sighting `getOwners()`/`getThreshold()` into per-event RPC. Good catch to call out explicitly so a future spec bump doesn't adopt them reflexively.

### Endorse

The "checked and rejected" section is the most valuable part — the timeseries rejection (running-total incompatibility + tiny claim volume) and the pruning-is-safe-because-Path-A-reads-archive-RPC reasoning are both correct and save a future reader from re-deriving them. Sequencing (graftable `networks.json` first, one clean-sync schema release second, with the immutable-flag-can't-graft constraint) is sound.

Accurate, actionable, and correctly scoped to "propose, don't unilaterally change the consumer API." Just tighten the #1 bound (or its claim) before acting on it.

PR #149, opened 2026-06-12
perf(pearl-transactions): clean-sync indexing release (#148 proposals)

rajat2502:
Implements the code-only proposals from **#148** (`INDEXING-PERFORMANCE.md`), skipping #1 (token start-blocks). This is the single **clean-sync release** the doc recommends — flipping `immutable` and changing entity shapes can't graft.

> Stacked on `feat/pearl-olas-sweep-category` (#147) so it keeps the `AGENT_OLAS_TO_MASTER` split, and **merges #145** (`fix/pearl-evicted-id-and-token-decimals`): with `FundsMovement` now immutable, the old eviction-id collision (`logIndex + i`) would be a **deterministic indexing halt** instead of a silent overwrite, so #145's two-segment id is a hard prerequisite of this release. Retarget to `main` once #147/#145 land.

## What's in it

**#2 — `TrackedAddress` merge.** Collapsed `TrackedSafe` / `TrackedEOA` / `StakingContract` lookups into one **immutable** `TrackedAddress` table (`role = MASTER | AGENT | MASTER_EOA | AGENT_EOA | STAKING`). `classifyTransfer`'s hot path now does **2 loads (from + to) instead of 6** with a fast exit for the ~99.99% untracked case; `getServiceRegistryAddress` resolves only in the dust branch; `getSrtuAddress` is memoized (it ran `Address.fromString` on every indexed transfer). Staking proxies get their `TrackedAddress(STAKING)` row in `handleInstanceCreated` — covered by a new producer→classifier wiring test.

**#3 — Immutable write-once entities.** `TrackedAddress`, `Token`, `StakingContract` → `@entity(immutable: true)`. graph-node skips block-range versioning → cheaper writes *and* cheaper reads.

**#4 — `BondMovement` split + `Service` Bytes id.** The only backfilled rows (SRTU bond deposit/refund) moved from `FundsMovement` into a small mutable `BondMovement`, letting the high-cardinality `FundsMovement` go **immutable**. `Service.id` is now **Bytes** (serviceId as Bytes), halving every `FundsMovement.service` FK.

**#5 — Spec bump + cleanups.** `specVersion 1.0.0 → 1.3.0`, `apiVersion 0.0.7 → 0.0.9` across the template **and the four committed per-network manifests** (the deploy workflow builds those as-is). No declared eth_calls — would turn the one-shot `getOwners`/`getThreshold` into per-event RPC. Removed redundant double-saves (pending-bond counter; the two `Service` saves). The `getOrCreateMasterSafe` re-sighting save is intentionally kept — it maintains `lastActivityTimestamp` (a real consumer field) and only fires on low-volume registry/staking events.

**Docs:** README + STEP5 queries rewritten for the union (`bondType` selected on `fundsMovements` would now be a GraphQL validation error); category table marks `SERVICE_BOND_*` as `BondMovement` rows; manifest `entities:` lists refreshed; the load-bearing write-order invariant in `getOrCreateAgentSafe` (Master Safe doubles as service operator) is documented in-code.

## ⚠️ Consumer-facing breaks (complete list — FE must update before deploy)
1. **Union required:** a complete ledger is `fundsMovements` **∪** `bondMovements` — bond rows are no longer in `fundsMovements`.
2. **`FundsMovement.bondType` is deleted** — any query selecting it hard-errors at GraphQL validation (the old README example did).
3. `SERVICE_BOND_DEPOSIT` / `SERVICE_BOND_REFUND` never appear as `FundsMovement.category` values.
4. `Service.id` is Bytes — read **`service.serviceId`** (numeric) for display, and any **id filters** (`where: { service: … }` on FundsMovement/BondMovement/DailyServiceFunds/AgentFundingEvent/AgentSafe) must use the `Bytes.fromBigInt` hex form (e.g. service 42 → `0x2a000000`), not `"42"`.
5. `TrackedSafe` / `TrackedEOA` entity types removed from the API; `TrackedAddress` added (internal-by-design, but previously queryable).
6. `AgentFundingEvent.id` derivation changed (serviceId bytes instead of UTF-8 string) — opaque ids, matters only if persisted client-side.
7. Unchanged, for the record: `FundsMovement.id` scheme and the `DailyServiceFunds.id` string format (`"42-<day>"`).

I'll do the FE union in lockstep with the deploy.

## Validation
- 48 Matchstick tests green (47 + #145's eviction-collision regression test); `codegen` + `build` clean on graph-cli 0.98.1 / graph-ts 0.38.2.
- Adversarial self-review (agent) ran against the #148 plan; all confirmed findings addressed in `530f25c` (manifests regenerated, README/STEP5 union queries, STAKING wiring test, SRTU memoization, invariant comment).
- Skipped #1 (token start-blocks) — needs the per-chain `MIN(firstSeenBlock)` query + a graft; deferred.

### DIFF
--- a/subgraphs/pearl-funds/IMPLEMENTATION-PLAN.md
+++ b/subgraphs/pearl-funds/IMPLEMENTATION-PLAN.md
@@ -1,9 +1,9 @@
 # Pearl Funds-Movement Subgraph — Implementation Plan
 
-**Status:** Implemented through Phase 2a (PRs #131/#132/#133 merged to `main`); Phase 2b (#138) in review. This plan is the design-of-record.
+**Status:** Implemented through Phase 2b (#138); clean-sync indexing release (#148 proposals / #149) consolidated the tracking + bond entities and flipped the high-cardinality ledger to immutable (see Rev. 8). This plan is the design-of-record.
 **Subgraph:** `subgraphs/pearl-transactions/` (renamed from `pearl-funds` per §11 #5)
 **Target networks (v1):** Gnosis, Polygon, Optimism, Base
-**Last updated:** 2026-06-01 (Rev. 7 — Phase 2b token-set reconciliation: pUSD is a separate Polymarket contract `0xC011a7E1…`, not a USDC.e UI alias; §4.5 token table + §4.3 networks table rebuilt to the shipped per-chain set (USDC / USDC.e / pUSD); §6.3 records the ship-on-chain decision + Polygon USDC.e rollback condition; Polystrat funds in USDC + pUSD. Rev. 6 — back-propagated the PR #131/#132 producer/consumer fixes so the plan matches shipped code: §3.3 SRTU is indexed, §4.6 SRTU-before-SR event order, §5.1/§5.2 inverted bond-attribution entities + handlers, §5.1 FundsMovement/Token mutability, §5.2 dual NFT guard + handleServiceStaked null-check, §5.4/§7/§8 staleness, pearl-transactions path. Rev. 5 — product decisions finalised: OPENING_BALANCE rows removed, opening balances delegated to the frontend via historyFloorBlock; native → Agent EOA confirmed accepted gap; AC #3 = Path A. Rev. 4 addressed PR #130 review + §11 #6. Rev. 3 added §4.5/§4.6. Rev. 2 added SRTU bond indexing, agent-ID anti-hardcoding, Master EOA tracking. Rev. 1 addressed PR #129 feedback.)
+**Last updated:** 2026-06-12 (Rev. 8 — back-propagated the clean-sync indexing release (#148 proposals #2/#3/#4/#5, PR #149) so the plan matches shipped code: `FundsMovement` flipped to `immutable: true` and the only mutated rows (SRTU bond deposit/refund) split into a new mutable `BondMovement` entity (§5.1) — the producer creates a `BondMovement` row and the consumer backfills it (§5.2), and a complete ledger is now `FundsMovement` ∪ `BondMovement`; `TrackedSafe` + `TrackedEOA` consolidated into a single immutable `TrackedAddress` (roles MASTER / AGENT / MASTER_EOA / AGENT_EOA / STAKING) so `classifyTransfer` does two hot-path loads instead of six (§6.4/§6.5); `Token` made immutable; service ids migrated to `Bytes` (`serviceEntityId`). Rev. 7 — Phase 2b token-set reconciliation: pUSD is a separate Polymarket contract `0xC011a7E1…`, not a USDC.e UI alias; §4.5 token table + §4.3 networks table rebuilt to the shipped per-chain set (USDC / USDC.e / pUSD); §6.3 records the ship-on-chain decision + Polygon USDC.e rollback condition; Polystrat funds in USDC + pUSD. Rev. 6 — back-propagated the PR #131/#132 producer/consumer fixes so the plan matches shipped code: §3.3 SRTU is indexed, §4.6 SRTU-before-SR event order, §5.1/§5.2 inverted bond-attribution entities + handlers, §5.1 FundsMovement/Token mutability, §5.2 dual NFT guard + handleServiceStaked null-check, §5.4/§7/§8 staleness, pearl-transactions path. Rev. 5 — product decisions finalised: OPENING_BALANCE rows removed, opening balances delegated to the frontend via historyFloorBlock; native → Agent EOA confirmed accepted gap; AC #3 = Path A. Rev. 4 addressed PR #130 review + §11 #6. Rev. 3 added §4.5/§4.6. Rev. 2 added SRTU bond indexing, agent-ID anti-hardcoding, Master EOA tracking. Rev. 1 addressed PR #129 feedback.)
 
 This document scopes a new subgraph that indexes **funds movement for the
 Master Safe and Agent Safe of Pearl predict services**. It covers Phase 1
@@ -110,8 +110,8 @@ review):
   new Pearl agent type launches** — the constant list lives in the
   compiled mapping, so any change requires redeploy + resync. This is
   exactly the kind of brittleness the trade subgraph plan avoided.
-- The indexing-cost concern in §2.2 is bounded by the size of
-  `TrackedSafe` in Phase 2, **not** by the total number of indexed
+- The indexing-cost concern in §2.2 is bounded by the size of the
+  `TrackedAddress` set in Phase 2, **not** by the total number of indexed
   services. `ServiceRegistryL2` event volume is low (same shape as the
   `service-registry` subgraph), so indexing every service is cheap.
 - Recording `agentIds` + `operators` on each `Service` preserves every
@@ -128,7 +128,7 @@ filter on these — they're consumer query parameters):
 | Gnosis (omenstrat) | **25** | Confirmed by maintainer; matches `valory-xyz/autonolas-subgraph` PR #89 (`PREDICT_AGENT_ID = 25`) |
 | Polygon (polystrat) | **86** | `predict-polymarket/src/constants.ts` |
 
-Phase 2's `TrackedSafe` set still needs a gate to keep the cost low. The
+Phase 2's `TrackedAddress` set still needs a gate to keep the cost low. The
 gate moves from "is this a Pearl predict service" to "does this service
 appear in the Pearl predict cohort or any other tracked cohort" — for
 v1 the only cohort we spawn `Safe` templates for is the Pearl predict
@@ -149,7 +149,7 @@ flows. Mode is intentionally omitted (deprecated network).
 Pearl predict services are the **primary consumer cohort** for v1
 (`agent ID 25` on Gnosis, `agent ID 86` on Polygon); the schema and
 data sources are deliberately agent-agnostic so other Pearl agent types
-(Optimus / babydegen / agents.fun) become drop-in additions of `TrackedSafe`
+(Optimus / babydegen / agents.fun) become drop-in additions of `TrackedAddress`
 seeding rather than full reindexes.
 
 ### 3.2 Phasing
@@ -169,7 +169,7 @@ visibility) and must clear both gates before commitment.
 
 - **Other Pearl agent types' Phase 2 cohorts** (Optimus / babydegen,
   agents.fun, etc.) — services for *every* agent type are still indexed
-  in Phase 1, but `TrackedSafe` seeding in Phase 2 is initially scoped to
+  in Phase 1, but `TrackedAddress` seeding in Phase 2 is initially scoped to
   the Pearl predict cohort to bound cost. Adding cohorts later is a
   per-deployment constant change, not a re-architecture.
 - **Mode network** — deprecated.
@@ -473,7 +473,7 @@ sequenceDiagram
 
 Bond-type attribution is best-effort per §5.2. Because the SRTU event
 fires *before* its ServiceRegistryL2 counterpart in every path, the SRTU
-handler is the **producer** (creates the `FundsMovement` row + enqueues
+handler is the **producer** (creates the `BondMovement` row + enqueues
 it) and the ServiceRegistryL2 handler is the **consumer** (`ActivateRegistration`
 / `RegisterInstance` / `TerminateService` / `OperatorUnbond` dequeue the
 row and backfill `serviceId` + `bondType`). The diagram shows the
@@ -614,18 +614,17 @@ enum FundsSource {
   RAW_TRANSFER                        # Direct ERC-20/native Transfer observed on chain
 }
 
-type FundsMovement @entity(immutable: false) {
-  # Mutable: SRTU TokenDeposit/TokenRefund rows are created amount-only by
-  # the SRTU producer and backfilled with serviceId + bondType by the
-  # ServiceRegistryL2 consumer later in the same tx (see §5.2). All other
-  # rows are write-once in practice.
+type FundsMovement @entity(immutable: true) {
+  # Immutable (Rev. 8): every FundsMovement is write-once. The only rows that
+  # were ever backfilled (SRTU TokenDeposit/TokenRefund bond rows) now live in
+  # the separate mutable BondMovement entity below, so this high-cardinality
+  # ledger skips block-range versioning — cheaper writes AND cheaper reads.
   id: Bytes!                          # txHash.concatI32(logIndex) — for semantic rows lacking a logIndex, use a stable sub-index
   service: Service
   masterSafe: MasterSafe
   agentSafe: AgentSafe
   category: FundsCategory!
   source: FundsSource!
-  bondType: ServiceBondType           # nullable; only populated for SERVICE_BOND_DEPOSIT / SERVICE_BOND_REFUND when disambiguation succeeds
   token: Bytes                        # token address (null for SAFE_DEPLOYED + pure NFT custody)
   amount: BigInt!                     # 0 for SAFE_DEPLOYED / SERVICE_EVICTED informational rows
   from: Bytes!
@@ -635,6 +634,29 @@ type FundsMovement @entity(immutable: false) {
   # Phase 2 backref — see §6.5; null on all Phase 1 rows and on Phase 2 rows
   # that aren't part of a multi-row agent-funding action.
   agentFundingEvent: AgentFundingEvent
+  blockNumber: BigInt!
+  blockTimestamp: BigInt!
+  transactionHash: Bytes!
+}
+
+# SRTU bond deposit / refund rows (Rev. 8). Split out of FundsMovement because
+# they are the ONLY rows that get mutated: the SRTU producer creates the row
+# serviceId/bondType-less and the matching ServiceRegistryL2 consumer backfills
+# serviceId + bondType + agentSafe later in the same tx (see PendingBondRow +
+# §5.2). Isolating this small mutable entity lets FundsMovement stay immutable.
+# Consumer note: a complete ledger is FundsMovement ∪ BondMovement.
+type BondMovement @entity(immutable: false) {
+  id: Bytes!                          # txHash.concatI32(logIndex)
+  service: Service
+  masterSafe: MasterSafe
+  agentSafe: AgentSafe
+  category: FundsCategory!            # SERVICE_BOND_DEPOSIT | SERVICE_BOND_REFUND
+  source: FundsSource!
+  bondType: ServiceBondType           # nullable; only populated when disambiguation succeeds
+  token: Bytes
+  amount: BigInt!
+  from: Bytes!
+  to: Bytes!
   blockNumber: BigInt!
   blockTimestamp: BigInt!
   transactionHash: Bytes!
@@ -672,7 +694,7 @@ type PendingRegistration @entity(immutable: false) {
 # Same-tx bond-attribution queue for SRTU TokenDeposit/TokenRefund
 # disambiguation. The SRTU event fires BEFORE its ServiceRegistryL2
 # counterpart in every path (§4.6), so the SRTU handler is the PRODUCER
-# (creates the FundsMovement row + enqueues its id) and the SR handler is
+# (creates the BondMovement row + enqueues its id) and the SR handler is
 # the CONSUMER (dequeues + backfills serviceId + bondType). Three entities:
 type PendingBondCounter @entity(immutable: false) {
   id: Bytes!                          # tx.hash
@@ -681,7 +703,7 @@ type PendingBondCounter @entity(immutable: false) {
 }
 type PendingBondRow @entity(immutable: false) {
   id: Bytes!                          # tx.hash.concatI32(slot)
-  fundsMovement: Bytes!               # FundsMovement id awaiting serviceId + bondType
+  bondMovement: Bytes!                # BondMovement id awaiting serviceId + bondType
   attributed: Boolean!
 }
 # Dedupe guard: registerAgents emits one TokenDeposit but RegisterInstance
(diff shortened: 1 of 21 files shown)


Tanya-atatakai:
## Code review — `perf/pearl-indexing-clean-sync` (automated multi-angle review)

7 finder angles × adversarial verification over the full branch diff vs `main`. 16 deduped candidates, 12 survived verification, top 10 below. Findings 1–2 are **new failure modes introduced by this branch** (consequences of `immutable: true` + the TrackedAddress table merge); 3–6 are pre-existing bugs in functions this branch rewrote; 7–10 are cleanups aligned with the branch's own perf goals.

**None of the findings below are covered by existing tests** — the bond-flow tests (`service-registry.test.ts`, `staking.test.ts`) have no `TokenBalance` assertions, `phase-2a.test.ts` seeds `TrackedAddress` roles directly via `seedTrackedAddress()` (bypassing the production `upsertTrackedAddress` write-order), and there is no coverage for master→master transfers, mixed-token funding events, or one address being both Agent Safe and Master Safe.

---

### 🔴 1. Duplicate `SafeTemplate.create` + immutable `FundsMovement` = deterministic indexing halt (new in this branch)

`SafeTemplate.create` is guarded by `MasterSafe.load` in one path (`src/utils.ts:189`) and `AgentSafe.load` in the other (`src/utils.ts:259`) — the guards are on different entities, so one address can get **two** template instances. Both instances then process the same `SafeReceived` log, and the second `handleSafeReceived` saves the same `tx.hash.concatI32(logIndex)` id. With `FundsMovement` now `immutable: true`, that second save is a deterministic "attempted to modify immutable entity" error that **permanently halts the subgraph**.

Reachable permissionlessly: transfer a service NFT to an existing Agent Safe (it's a Gnosis Safe, so it passes the `getOwners()` probe in `handleServiceNftTransfer`) → `getOrCreateMasterSafe` spawns a second template on that address. On `main` this was a silent idempotent overwrite.

**Fix:** gate `SafeTemplate.create` on a shared per-address sentinel (check both `MasterSafe.load` and `AgentSafe.load`, or a dedicated marker entity).

### 🔴 2. Write-once `TrackedAddress` role poisoning (regression vs the two-table design)

`upsertTrackedAddress` early-returns on any existing row with no role priority (`src/utils.ts:617-618`), and the operator loop (`src/utils.ts:260-269`) writes **every operator of every indexed service** — including masterless/non-Pearl services — as `AGENT_EOA`. If a Safe is recorded as operator of a non-Pearl/EOA-owned service first, a later `MASTER` upsert for that address silently no-ops; `classifyTransfer` then routes every transfer for that Master Safe to `OTHER` with no `masterSafe` ref, and `linkBondMasterSafe`'s `role == "MASTER"` guard fails — the user's entire wallet history is permanently empty, unrecoverable without a resync (entity is `immutable: true`).

On `main`, the TrackedSafe/TrackedEOA tables let both roles coexist and the Safe role won. The "LOAD-BEARING write-order" comment (`src/utils.ts:244`) only covers the Pearl happy path, not the permissionless-registry collision.

**Fix:** skip operators equal to `service.masterSafe` in the loop, and/or give `upsertTrackedAddress` explicit role-priority semantics (MASTER wins).

### 🟠 3. Master↔SRTU suppression skips `TokenBalance` — and the rewritten comment claims the opposite

The comment at `src/utils.ts:758-759` ("The OLAS TokenBalance delta is updated separately in handleErc20Transfer and is unaffected") is factually false: `handleErc20Transfer` returns on `null` classification (`src/erc20.ts:41-44`) **before** every `upsertTokenBalance` call, and `upsertTokenBalance` has no other call sites. So bond deposits/refunds never touch `TokenBalance` — a Master Safe's OLAS balance is overstated by the full bonded amount for the entire staking period. Pre-existing on `main`, but this diff deletes and re-writes the comment block, re-asserting the false invariant. Either apply the delta before the `null` return or fix the comment to document the gap.

### 🟠 4. Master→Master transfers: recipient credited, sender never debited

In the rewritten `classifyTransfer`, `toRole == "MASTER"` (`src/utils.ts:798`) wins over `fromRole == "MASTER"` (`:815`), so a Master Safe A → Master Safe B transfer classifies only as `MASTER_FUNDING_IN(masterSafe=B)`. The balance branches in `erc20.ts` never debit A, and A's masterSafe-filtered history shows no outgoing row. Realistic for funds migration between Pearl installs. Pre-existing branch order, but the function was rewritten here and there's no MASTER→MASTER test.

### 🟠 5. `AgentFundingEvent.totalOlasAmount` sums mixed-decimal tokens

`handleErc20Transfer` is shared by OLAS, wrapped-native, and stablecoin data sources, and the MASTER_TO_AGENT block calls `addToAgentFundingEvent(afe, amount, false)` for every token (`src/erc20.ts:101`). The grouping key is per-(tx, masterSafe, service) — not per-token — so one tx sending 1 OLAS (1e18) + 100 USDC (1e8) yields `totalOlasAmount = 1000000000100000000`, contradicting the field's own doc ("sum across constituent OLAS rows", `schema.graphql:288`). Pre-existing, but sits directly under this branch's AGENT_OLAS_TO_MASTER re-tag.

### 🟡 6. Shared agent EOA's `service` attribution is now frozen forever

`TrackedAddress(AGENT_EOA)` is created with `service` = the first service that saw that operator (`src/utils.ts:260-269`); with the entity immutable and the upsert write-once, an operator EOA serving a second service has all its MASTER_TO_AGENT rows attributed to service #1 permanently. `main` had the same first-write-wins behavior, but its mutable `TrackedEOA` left a fix-up path that `immutable: true` forecloses. Low impact for masterSafe-filtered Pearl queries; worth a conscious sign-off.

---

### Cleanups (all confirmed, all aligned with this branch's perf goals)

7. **Magic strings vs constants** — the rewritten `classifyTransfer` returns raw literals (`"MASTER_TO_AGENT"`, `"MASTER_FUNDING_IN"`, … `src/utils.ts:766-830`) while `constants.ts` exports a matching `CATEGORY_*` for each (the same function already uses `CATEGORY_OTHER`); `erc20.ts:113/125` compares the literal `"AGENT_OLAS_TO_MASTER"` three lines below importing the constant; and the new role strings (`"MASTER"`, `"AGENT_EOA"`, `"STAKING"`, …) have no `ROLE_*` constants at all. A typo compiles silently and produces permanent silent misclassification (roles are write-once).
8. **Avoidable hot-path load** — in the `toRole == "MASTER"` branch, `MasterSafe.load` (`src/utils.ts:806`) runs on every inbound transfer but its result is only consumed when `fromRole == "MASTER_EOA"`. Checking `fromRole` first is behavior-preserving and drops a `store.get` from every ordinary deposit — the exact metric this branch optimizes.
9. **Double save** — `getOrCreateAgentFundingEvent` saves the fresh entity (`src/utils.ts:885`) and both call sites immediately call `addToAgentFundingEvent`, which saves again. This diff applied exactly this fix to `getOrCreatePendingBondCounter` ("No save here — the only caller saves after bumping", `src/utils.ts:445`) but missed the identical pattern in a function it touched.
10. **Dead load in a no-op** — `emitNativeOutPlaceholder` (`src/safe.ts:108`) still executes `MasterSafe.load` on every `ExecutionSuccess`/`ExecutionFromModuleSuccess` of every tracked Safe, and the diff updated its comments (TrackedSafe→TrackedAddress) instead of emptying the dead body; the surviving comment block is internally contradictory.

<sub>Honorable mentions cut by the 10-finding cap: `serviceEntityId` is the 4th byte-identical copy of `Bytes.fromByteArray(Bytes.fromBigInt(serviceId))` in `utils.ts` (+3 independent re-derivations in tests); the hand-rolled `srtuCacheNetwork`/`srtuCacheAddr` memo could be zero-mutable-state module-init constants.</sub>

<sub>Verified-and-refuted (for the record): the `bondType`→`BondMovement` split, `Service.id` → Bytes, and the `AGENT_OLAS_TO_MASTER` re-tag are all documented, deliberate, and shielded by version-pinned Studio endpoints; the UNKNOWN/6-decimals token fallback is safe because `log.critical` halts indexing before the bad save.</sub>

Tanya-atatakai:
## Missing-coverage repros for review findings 2–5

Follow-up to the [review comment](https://github.com/valory-xyz/autonolas-subgraph-studio/pull/149#issuecomment-4692072698): none of findings 2–5 are exercised by the current suites, largely because `phase-2a.test.ts` seeds `TrackedAddress` roles via its own `seedTrackedAddress()` helper instead of the production write path, and the bond-flow suites contain no `TokenBalance` assertions at all. Below are four drop-in tests using the suites' existing fixtures/helpers. **Each fails today** — they encode the expected behavior, not the current one — so they're repro cases to pair with the fixes (or, where a behavior is consciously accepted, to invert into pinning tests with a comment).

### Finding 2 — write-once role poisoning (`tests/service-registry.test.ts`)

Drives the production write order (operator loop before MASTER discovery) instead of seeding roles directly. Uses the suite's existing `newRegisterInstance` / `newCreateMultisig` / `newNftTransfer` / `mockGetOwners` / `mockGetThreshold` helpers.

```ts
test("a Safe first seen as operator of a masterless service can still become a Master Safe", () => {
  // An address that operates a non-Pearl service today and becomes a
  // Pearl Master Safe tomorrow. The registry is permissionless, so the
  // operator loop in getOrCreateAgentSafe sees it first.
  const POISONED_SAFE = Address.fromString(
    "0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee"
  );
  const OTHER_MULTISIG = Address.fromString(
    "0xffffffffffffffffffffffffffffffffffffffff"
  );

  // 1. Masterless service (no NFT mint indexed to a Safe): operator-first.
  const tx1 = mockTx(701);
  handleRegisterInstance(
    newRegisterInstance(POISONED_SAFE, SERVICE_ID, AGENT_INSTANCE_1, BigInt.fromI32(1), tx1, 0)
  );
  handleCreateMultisigWithAgents(
    newCreateMultisig(SERVICE_ID, OTHER_MULTISIG, tx1, 1)
  );
  // Sanity: the operator loop wrote the poisoned row.
  assert.fieldEquals(
    "TrackedAddress", POISONED_SAFE.toHexString(), "role", "AGENT_EOA"
  );

  // 2. The same Safe later onboards Pearl: service #2's NFT mints to it.
  mockGetOwners(POISONED_SAFE, [MASTER_EOA, BACKUP_EOA]);
  mockGetThreshold(POISONED_SAFE, 1);
  const tx2 = mockTx(702);
  handleServiceNftTransfer(
    newNftTransfer(Address.zero(), POISONED_SAFE, OTHER_SERVICE_ID, tx2, 0)
  );

  // The MasterSafe entity is created…
  assert.fieldEquals(
    "MasterSafe", POISONED_SAFE.toHexString(), "setupTransferSeen", "false"
  );
  // …but upsertTrackedAddress early-returned on the existing AGENT_EOA row.
  // FAILS today: role stays "AGENT_EOA", so classifyTransfer routes every
  // transfer for this Master Safe to OTHER forever (TrackedAddress is
  // immutable — no recovery without a resync).
  assert.fieldEquals(
    "TrackedAddress", POISONED_SAFE.toHexString(), "role", "MASTER"
  );
});
```

### Finding 3 — Master↔SRTU suppression skips the TokenBalance delta (`tests/phase-2a.test.ts`)

```ts
test("Master Safe → SRTU bond leg debits the OLAS TokenBalance", () => {
  seedMasterSafe(/* setupTransferSeen = */ true);
  const SRTU = Address.fromString(
    "0xa45E64d13A30a51b91ae0eb182e88a40e9b18eD8" // SRTU_GNOSIS
  );
  const balId = MASTER_SAFE.concat(OLAS_GNOSIS);

  // Fund the Master Safe with 5 OLAS.
  const tx1 = mockTx(801);
  handleErc20Transfer(newOlasTransfer(MASTER_EOA, MASTER_SAFE, AMOUNT, tx1, 0));
  assert.fieldEquals("TokenBalance", balId.toHexString(), "balance", AMOUNT.toString());

  // Post a 1 OLAS bond: Master Safe → SRTU.
  const bond = BigInt.fromString("1000000000000000000");
  const tx2 = mockTx(802);
  handleErc20Transfer(newOlasTransfer(MASTER_SAFE, SRTU, bond, tx2, 0));

  // The RAW_TRANSFER row is correctly suppressed (deduped against the
  // SEMANTIC bond row)…
  assert.notInStore("FundsMovement", tx2.concatI32(0).toHexString());
  // …but the comment at utils.ts:758 claims the TokenBalance delta still
  // lands "separately in handleErc20Transfer". FAILS today: the null
  // classification returns before any upsertTokenBalance call, so the
  // balance stays 5 OLAS and overstates the wallet by the bonded amount
  // for the whole staking period.
  assert.fieldEquals(
    "TokenBalance", balId.toHexString(), "balance", AMOUNT.minus(bond).toString()
  );
});
```

### Finding 4 — Master Safe → Master Safe never debits the sender (`tests/phase-2a.test.ts`)

```ts
test("Master Safe A → Master Safe B debits A's TokenBalance", () => {
  seedMasterSafe(true); // A = MASTER_SAFE

  // Seed a second, independent Master Safe B (e.g. a new Pearl install).
  const MASTER_B = Address.fromString(
    "0x9999999999999999999999999999999999999999"
  );
  const msB = new MasterSafe(MASTER_B);
  msB.network = "gnosis";
  msB.masterEoa = NEW_OWNER;
  msB.owners = [NEW_OWNER];
  msB.threshold = BigInt.fromI32(1);
  msB.firstSeenTimestamp = BigInt.fromI32(1);
  msB.firstSeenBlock = BigInt.fromI32(1);
  msB.historyFloorBlock = BigInt.fromI32(1);
  msB.historyFloorTimestamp = BigInt.fromI32(1);
  msB.lastActivityTimestamp = BigInt.fromI32(1);
  msB.setupTransferSeen = true;
  msB.save();
  seedTrackedAddress(MASTER_B, "MASTER", MASTER_B, null);

  // Fund A with 5 OLAS, then migrate all of it to B.
  const tx1 = mockTx(901);
  handleErc20Transfer(newOlasTransfer(MASTER_EOA, MASTER_SAFE, AMOUNT, tx1, 0));
  const tx2 = mockTx(902);
  handleErc20Transfer(newOlasTransfer(MASTER_SAFE, MASTER_B, AMOUNT, tx2, 0));

  // B is credited (the toRole == MASTER branch wins) — passes today.
  assert.fieldEquals(
    "TokenBalance",
    MASTER_B.concat(OLAS_GNOSIS).toHexString(),
    "balance",
    AMOUNT.toString()
  );
  // FAILS today: the single MASTER_FUNDING_IN classification never reaches
  // the MASTER_WITHDRAWAL debit branch, so A's balance stays 5 OLAS forever
  // (and A's masterSafe-filtered history shows no outgoing row — the tx2
  // row's masterSafe is B).
  assert.fieldEquals(
    "TokenBalance",
    MASTER_SAFE.concat(OLAS_GNOSIS).toHexString(),
    "balance",
    "0"
  );
});
```

### Finding 5 — `totalOlasAmount` sums mixed-decimal tokens (`tests/phase-2a.test.ts`)

```ts
test("same-tx OLAS + USDC Master → Agent funding keeps totalOlasAmount OLAS-only", () => {
  seedMasterSafe(true);
  seedAgentSafe(SERVICE_ID);
  const USDC = Address.fromString(
    "0xDDAfbb505ad214D7b80b1f830fcCc89B60fb7A83" // gnosis USDC, 6 dec
  );

  const tx = mockTx(1001);
  const olasLeg = BigInt.fromString("1000000000000000000"); // 1 OLAS (1e18)
  const usdcLeg = BigInt.fromString("100000000");           // 100 USDC (1e8)
  handleErc20Transfer(newOlasTransfer(MASTER_SAFE, AGENT_SAFE, olasLeg, tx, 0));
  handleErc20Transfer(
    newErc20TransferAt(USDC, MASTER_SAFE, AGENT_SAFE, usdcLeg, tx, 1)
  );

  // Both legs group under one event (per-(tx, masterSafe, service) key).
  assert.entityCount("AgentFundingEvent", 1);
  const afeId = tx.concat(MASTER_SAFE).concat(serviceBytes(SERVICE_ID));
  // schema.graphql documents totalOlasAmount as "sum across constituent
  // OLAS rows". FAILS today: actual value is 1000000000100000000 —
  // 1e18 OLAS + 1e8 USDC raw units summed across decimals, which renders
  // as ~1.0000000001 OLAS and silently drops the USDC leg.
  assert.fieldEquals(
    "AgentFundingEvent", afeId.toHexString(), "totalOlasAmount", olasLeg.toString()
  );
});
```

### Notes

- Finding 2's test needs `handleRegisterInstance` / `handleServiceNftTransfer` (already imported in `service-registry.test.ts`) plus `mockGetThreshold` for the probe; everything else reuses existing fixtures.
- Finding 5's fix has a design choice baked into the assertion: either gate `addToAgentFundingEvent(…, false)` on `event.address == getOlasAddress()` (keeping the field's documented OLAS-only meaning, as asserted above), or rename/restructure the field for multi-token sums — in which case adjust the expected value accordingly.
- Finding 1 (duplicate `SafeTemplate.create` → immutable `FundsMovement` halt) is deliberately not included: Matchstick records `dataSource.create` calls but doesn't simulate two template instances processing the same log, so it can't reproduce the halt. That one needs the code-level guard, plus at most an assertion that `SafeTemplate.create` isn't reachable twice for one address.

rajat2502:
Thanks @Tanya-atatakai — exceptionally useful review. All 10 findings addressed in `1da100d`; the four repro tests are in (adapted to assert the fixed behavior).

**🔴 F1 — duplicate template halt: fixed.** `SafeTemplate.create` is now cross-guarded — `getOrCreateMasterSafe` skips when an `AgentSafe` exists at the address, and vice-versa, so one address gets exactly one template. Test drives the NFT-to-existing-Agent-Safe scenario; as you predicted, matchstick can't observe a duplicate create on the *same* address (it counts unique created data sources — verified with a probe), so the test asserts the both-tables collision state and the guard is the protection.

**🔴 F2 — role poisoning: fixed.** Operator rows are now written **only for Pearl-linked services** (`service.masterSafe` set), and the master safe itself is skipped in the loop — a masterless/non-Pearl service can no longer pre-claim an address that later becomes a Pearl Master Safe. Your repro is in (adjusted: after the fix the masterless step writes *no* row, then the NFT mint lands `MASTER`). This also retires the fragile "write-order invariant" comment — the loop no longer depends on it.

**🟠 F3 — bond TokenBalance: fixed properly** (not just the comment). The SRTU handlers now own the bond's balance delta — `handleTokenDeposit` debits / `handleTokenRefund` credits the tracked Master Safe, exactly once (the raw leg stays suppressed). The false comment in `classifyTransfer` is corrected to say exactly that. Your repro is in, extended to fire both same-tx events as on-chain.

**🟠 F4 — master→master: sender now debited.** New `ClassifyResult.senderMasterId` (set only on MASTER→MASTER) lets `handleErc20Transfer` debit the sender with zero extra store loads. The sender's masterSafe-filtered *history* intentionally still shows no outgoing row — documented in CLAUDE.md "Honest limits" as the same accepted shape as the native-out gap. Repro in (both-sides balance asserted).

**🟠 F5 — totalOlasAmount: OLAS-gated.** The bump now fires only when `event.address == OLAS`; stablecoin legs still link via `transfers`. Took the first of your two design options (keep the documented OLAS-only meaning). Repro in.

**🟡 F6 — accepted + documented** on `upsertTrackedAddress` (first-write-wins shared-operator attribution; harmless for masterSafe-filtered queries).

**Cleanups 7–10: all done.** `CATEGORY_*` constants throughout `classifyTransfer`/handlers + new `ROLE_*` constants (a typo in a write-once role would be permanent); `MasterSafe.load` moved behind the `fromRole == MASTER_EOA` check (one `store.get` off every ordinary deposit); `getOrCreateAgentFundingEvent` double-save removed; `handleSafeExecution*` are true no-ops now (dead `MasterSafe.load` per execution removed). Honorable mention too: `serviceEntityId` is now the single source of the serviceId byte encoding (others alias it).

**Validation:** 53/53 Matchstick tests (48 + your 4 repros + the poisoning production-write-order test); codegen + build clean.

Tanya-atatakai (dismissed):
## Approve with nits ✅

Solid, defensively-reasoned clean-sync refactor. Traced every signature change across call sites, every TokenBalance leg for double-count/missing-leg, and the immutable-entity flips — the headline correctness risks are clean:

- No double-count on bonds / Master→Master migration / `MASTER_TO_AGENT` (the dedicated `to`-credit block is correct precisely because `MASTER_TO_AGENT` is omitted from the credit list).
- All call sites consistently migrated to `serviceEntityId` (Bytes ids) / `BondMovement` / `bondMovement`.
- SRTU cold-start debit fires — the service-NFT mint to the Master Safe precedes the first security-deposit `TokenDeposit` in the same multicall, so `isTrackedMaster` resolves.
- Cross-tenant pre-claim vector genuinely closed (operator rows written only for Pearl-linked services).
- `ServicesEvicted` id-collision fix is correct.

### Non-blocking nits

1. **Role poisoning — `src/utils.ts:189`.** An address already tracked `AGENT` that later receives a service NFT keeps role `AGENT` forever (the immutable `MASTER` write early-returns), so `MASTER_FUNDING_IN`/`SAFE_SETUP_TRANSFER` never fire for it and that install's history is mis-bucketed. This is the exact case the duplicate-template guard comment anticipates — the guard prevents the double-spawn crash but doesn't reconcile the role. Accept+document, or promote the role when a `MasterSafe` entity is created for an already-tracked address.
2. **SRTU manifest under-declares written entities — `subgraph.template.yaml:63` (+4 generated).** `handleTokenDeposit`/`handleTokenRefund` now write `Token` + `TokenBalance` via `upsertTokenBalance`, but the SRTU `entities:` list is only `BondMovement / PendingBondCounter / PendingBondRow`. Advisory in current graph-node, but a convention break vs every other source.
3. **No test for the SRTU `TokenBalance` debit/credit under real ordering.** The debit gate silently no-ops if MASTER isn't tracked yet, with no fallback; existing bond tests assert only `BondMovement`, so the balance path is unguarded. Add a mint→deposit→assert-debited test.
4. **Hot-path resolvers left unmemoized — `src/constants.ts:57,125`.** `getSrtuAddress` was memoized this PR on a millions-of-events rationale, but `getOlasAddress` (called 2× per OLAS event at `erc20.ts:61,112`) and `getServiceRegistryAddress` weren't. Same cache pattern or hoist into a local.
5. **Docs drift.** `CLAUDE.md` still describes the removed `TrackedSafe`/`TrackedEOA` + mutable `FundsMovement`; test-count references (53/46) are stale (suite runs 48).

Native-path `TokenBalance` no-op and the `serviceIndexId` aliasing / no-op execution handlers are accepted as intentional.

Tanya-atatakai (approved):

mohandast52 (approved):

PR #151, opened 2026-06-15
fix(predict): sync omen/polymarket fixes + answerFinalizedTimestamp + head-to-head markets

### LINKED ISSUE #114: predict subgraph: expose answerFinalizedTimestamp on FixedProductMarketMakerCreation (blocks trader#903 / ZD#919)
## Problem

The `predict` Omen subgraph (currently served from `https://api.subgraph.staging.autonolas.tech/api/proxy/predict-omen`, since the gateway URL is stuck) exposes `currentAnswer` and `currentAnswerTimestamp` on `FixedProductMarketMakerCreation`, but not `answerFinalizedTimestamp`.

Reality.eth answers can flip during the 24-hour dispute window. Without a finalization field, consumers can't distinguish a provisional answer from a settled one. This breaks any downstream logic that needs to gate user-visible status on finalization — notably the "invalid market" label (ZD#919): a temporary `0xff…ff` sentinel gets cached as permanently invalid even after it's overwritten on-chain.

## Evidence

Introspection on the live endpoint:

```
$ curl -sS -X POST -H 'Content-Type: application/json' \
  -d '{"query":"{ __type(name:\"FixedProductMarketMakerCreation\") { fields { name } } }"}' \
  https://api.subgraph.staging.autonolas.tech/api/proxy/predict-omen
```

Returned fields: `id, creator, conditionalTokens, collateralToken, conditionIds, question, outcomes, fee, currentAnswer, currentAnswerTimestamp, bets, participants, blockNumber, blockTimestamp, transactionHash`. No `answerFinalizedTimestamp`.

Attempting to select it produces:

```
Type `FixedProductMarketMakerCreation` has no field `answerFinalizedTimestamp`
```

## Ask

Add `answerFinalizedTimestamp: BigInt` to the `FixedProductMarketMakerCreation` entity, populated from Realitio's `LogFinalize(bytes32 question_id, bytes32 answer)` event (`block.timestamp` at finalization). Semantics should match Protofire's `protofire/omen-xdai` subgraph, which exposes the same field on its `FixedProductMarketMaker` entity.

## Blocked consumer

`valory-xyz/trader` PR #903 — three queries in `packages/valory/skills/agent_performance_summary_abci/graph_tooling/queries.py` need this field:

- `GET_PREDICTION_HISTORY_QUERY`
- `GET_TRADER_AGENT_BETS_QUERY`
- `GET_SPECIFIC_MARKET_BETS_QUERY`

The PR is functionally complete on the consumer side, and is held pending this schema addition. Unit tests pass on fabricated fixtures, but any live query 500s until the field exists.

## Nice-to-have

A CI smoke test (introspection or a canary query) on the subgraph repo would have caught this mismatch before consumers hit it.

### LINKED ISSUE #128: predict-polymarket: Question/MarketMetadata not created for head-to-head (non-Yes/No) markets → null bet.question
## Summary

In the **predict-polymarket** subgraph (Polygon; served at `api.subgraph.autonolas.tech/api/proxy/predict-polymarket`), `bet.question` (and `marketParticipant.question`) resolve to **`null`** for a class of Polymarket binary markets. The subgraph indexes the condition and tokens for these markets but **never creates the `Question` / `MarketMetadata` entity**, even though all the metadata exists on-chain.

This is a **mapping/coverage gap**, not on-chain reality.

> **Status (root-caused from source):** Confirmed. The handler deliberately drops any binary market whose two outcomes aren't literally `Yes`/`No`. See [Root cause](#root-cause-confirmed-from-mapping-source) and [Fix](#fix) below.

## Reproduction

```graphql
{
  bets(where: { question: null }, orderBy: blockTimestamp, orderDirection: desc) {
    id
    blockTimestamp
    outcomeIndex
    bettor { id }
    question { id }   # null
  }
}
```

## Affected market class

All affected markets are **head-to-head e-sports / match-winner markets** where the two outcomes are **team/competitor names, not Yes/No**, e.g.:

- `Dota 2: Team Nemesis vs Zero Tenacity (BO3) - 1win Essence Group B`
- `LoL: Gen.G vs Kiwoom DRX (BO3) - LCK Rounds 1-2`
- `Counter-Strike: FUT Esports vs Astralis (BO3) - BLAST Rivals Group A`
- `Honor of Kings: SYGaming vs EDward Gaming (BO7) - Honor of Kings Challenger Cup`
- `Internazionali BNL d'Italia (tennis): Pablo Carreno Busta vs Stan Wawrinka`

The subgraph **does** correctly index standard **Yes/No** markets, including Yes/No *about* sports/e-sports (e.g. "Will GAM Esports win LCP 2026?", "…Both Teams to Score"). Only the head-to-head "`<Sport>: A vs B (BOx) - <Tournament>`" template (team-name outcomes) is missing a `Question`.

## Evidence: on-chain vs subgraph

Null market (e-sports) vs a working market (politics), side by side:

| | NULL market (e-sports) | LINKED market (politics) |
|---|---|---|
| condition | `0xe749a3774e36c0add2c61c5e8cbae419823daea76d513067fe93f60327c40ec1` | `0x70d706fe…5d26de` |
| on-chain `ConditionPreparation` | ✅ block 86430029, oracle `0x65070be9…` (UmaCtfAdapter), 2 slots | ✅ oracle `0x65070be9…` (UmaCtfAdapter), 2 slots |
| subgraph `QuestionIdToConditionId` | ✅ present | ✅ present |
| subgraph `TokenRegistry` | ✅ 2 entries | ✅ 2 entries |
| **subgraph `Question` entity** | ❌ **MISSING** | ✅ present (title *"US x Iran diplomatic meeting by May 15, 2026?"*) |
| on-chain title / ancillary | ✅ present | ✅ present |

Worked example (questionId `0x667d1ca6e5479abf9f8b4b3e0c67e69665d7d1c0a2703357f67bff8c8555d8e0`, condition `0xe749a377…c40ec1`):
- Prepared by the **standard** `UmaCtfAdapter` `0x65070be91477460d8a7aeeb94ef92fe056c2f2a7` — the *same* adapter as working Yes/No markets.
- Binary (2 outcome slots). Not neg-risk (no NegRisk adapter/exchange involved).
- Init tx `0xf9168165a3e49b3aaf7f97d9d563915ce1cb77da265b8aaef2a4679bae85e487` emits the full ancillary from the adapter:
  > YES_OR_NO_QUERY … q: title: **Dota 2: Team Nemesis vs Zero Tenacity (BO3) - 1win Essence Group B**, description: This market refers to the Dota 2 match …
- Compare a working market's init tx `0xae856f7ce1c84d30b819fef30559aadc69d2b09df8835625d87cb18a325e4259` ("US x Iran diplomatic meeting…") which **does** get a `Question`.

All **18 distinct null conditions** behave identically: every one prepared by the same `UmaCtfAdapter 0x65070be9…`, binary, `Question` missing (18/18).

## Root cause (confirmed from mapping source)

The guard is in [`subgraphs/predict/predict-polymarket/src/uma-mapping.ts`](https://github.com/valory-xyz/autonolas-subgraph-studio/blob/main/subgraphs/predict/predict-polymarket/src/uma-mapping.ts). `extractBinaryOutcomes()` extracts the two outcome labels from the ancillary's `p1 corresponds to … , p2 to …` mapping, then **throws them away unless they're literally Yes/No**:

```ts
// uma-mapping.ts:113-117 — the p1/p2 mapping path
if (isYesNoPair(out1, out2)) {
  return [out1, out2];
} else {
  return []; // Reject categorical markets   ← head-to-head dies here
}
```

The head-to-head ancillary carries `…Where p1 corresponds to Team Nemesis, p2 to Zero Tenacity, p3 to unknown/50-50`, so `isYesNoPair` is false and the function returns `[]`. The handler then early-returns before creating anything:

```ts
// uma-mapping.ts:149-152
let outcomes = extractBinaryOutcomes(rawData);
if (outcomes.length == 0) return;   // ← no MarketMetadata, no Question
```

The bug is **conflating "non-Yes/No labels" with "categorical / >2 outcomes."** These markets are genuinely binary (2 outcome slots) — they just use team-name labels instead of Yes/No. The labels are purely descriptive: nothing downstream branches on the strings. The winning side comes from `payouts`/`winningOutcome`, and `bet.outcomeIndex` (0/1) comes from `TokenRegistry` (derived from the index set in `conditional-tokens.ts`), both independent of the labels.

### Answers to the open questions

1. **Which dataSource, and why Iran works but Dota doesn't.** Both go through the *same* path. `Question`/`MarketMetadata` is created by `QuestionInitialized` on the `OptimisticOracleV3` dataSource — manifest address `0x65070BE9…` ([`subgraph.yaml`](https://github.com/valory-xyz/autonolas-subgraph-studio/blob/main/subgraphs/predict/predict-polymarket/subgraph.yaml), the address this issue calls "UmaCtfAdapter"), routed `handleOOQuestionInitialized → handleQuestionInitialization`. The event **is** received for the Dota market; it's dropped solely by the Yes/No content guard above. (Naming note: the manifest's `UmaCtfAdapter` *name* is on a different address, `0x157Ce2…` — a label mismatch, not a functional one.)
2. **Missing second adapter/registry?** No. The oracle `0x65070BE9…` is in the manifest with the correct `startBlock`. Nothing is missing from `dataSources`; the gap is 100% the mapping guard.
3. **Re-index / backfill required?** **Yes — and a forward-only deploy repairs nothing.** The 18 conditions' `QuestionInitialized` events already fired in the past and won't fire again, so shipping the mapping fix to the live deployment does zero for existing data. There's also an asymmetry between the two null relations:
   - `MarketParticipant.question` stores the raw `conditionId` (`utils.ts:129`), so it would **auto-resolve** the moment a matching `Question` exists.
   - `bet.question` is written only at bet-time, inside `if (question !== null)` (`ctf-exchange.ts:100-103`, same in `ctf-exchange-v2.ts:64-66`) — so it stays `null` forever unless the **bet's `OrderFilled` is re-processed**.

   Therefore the re-index must **begin before the earliest affected `ConditionPreparation`** (which precedes the trades), so that both the `Question` is created *and* the bets are re-processed against it. A **full re-sync** from the original `startBlock` (78425180) is the safe option; a graft just before the earliest affected condition (~block 86430029) is faster but only safe if no head-to-head markets exist earlier in history (not yet verified). `Question` is `@entity(immutable: true)`, so it cannot be patched in place.

### Secondary effect (profit aggregates)

`processMarketResolution` does `let question = Question.load(conditionId); if (question == null) return;` (`utils.ts:196-197`). So a missing `Question` doesn't just null `bet.question` — it means **resolution is never processed** for these bets, which is exactly why they carry `countedInTotal` / `countedInProfit = false` and are excluded from the subgraph's own profit aggregates. The re-sync fixes both symptoms.

> ⚠️ **Expected aggregate shift after re-sync.** Once the Questions exist, `processMarketResolution` stops short-circuiting and folds these 18 conditions into the `Global` / `TraderAgent` / `DailyProfitStatistic` totals that previously excluded them. Published ROI/profit numbers will therefore **change retroactively** — this is the correction, not a new regression. Polystrat / olas-website owners should be told to expect the one-time shift.

## Fix

Relax `extractBinaryOutcomes` to accept any **2-outcome** market, preserving the real team-name labels, while still rejecting genuinely multi-outcome markets (e.g. `outcomes: [Yes, No, Maybe]`). Binary-ness is already guaranteed upstream: `handleConditionPreparation` only creates the `QuestionIdToConditionId` bridge when `outcomeSlotCount == 2` (`conditional-tokens.ts:68`), and `handleQuestionInitialization` bails when the bridge is missing.

- Mapping change in `uma-mapping.ts`: return `[out1, out2]` from the p1/p2 path regardless of Yes/No; return the two labels from the `outcomes: [A, B]` path when the list length is exactly 2.
- Three existing tests (`tests/uma-mapping.test.ts:113`, `:184`, `:273`) encode the *buggy* expectation (assert team/Option markets are rejected) and flip to asserting these markets are now created with their real labels. The 3-outcome rejection test (`:147`) stays.
- **Harden the label parser.** The p1/p2 extractor truncates `out2` at the first `,`, falling back to the first `.` (`uma-mapping.ts:106`). That fallback was harmless while these markets were rejected, but now that we surface the labels it can clip team names containing a period (`Gen.G`, `St. Louis`). Add coverage for period/comma-bearing names so we don't ship slightly-wrong outcome labels. (Severity is low — labels are descriptive only; no profit/index logic reads them — but the whole point of the fix is correct team names.)
- **Deploy must be a full re-sync** to retro-link the already-stored null bets (see answer 3 above) — a mapping fix alone won't repair existing data.

## Impact / scope

- Window observed: **2026-04-30 → 2026-05-06** — 49 null bets, 18 conditions, 1 bettor (`0x9070a95134a06290141e94b6dd453c9c88dc4ba4`). Currently dormant only because that agent stopped betting head-to-head e-sports after 05-06; **latent and will recur** for any tracked agent betting match-winner markets.
- Downstream: this silently broke the olas-website Polystrat tool-accuracy table (went empty) and froze Polystrat ROI. olas-website [PR #506](https://github.com/valory-xyz/olas-website/pull/506) added null-guards, but that **only masks the symptom** — these bets keep getting dropped from accuracy/ROI until the subgraph creates the `Question` entities.
- Secondary: these null bets also have `countedInTotal` / `countedInProfit = false`, so they're excluded from the subgraph's own profit aggregates as well.

---
*Root-caused via on-chain + subgraph forensics (Polygon RPC + live subgraph), confirmed against the mapping source.*

Tanya-atatakai:
## Summary

Syncs the predict subgraph work from the [`autonolas-subgraph`](https://github.com/valory-xyz/autonolas-subgraph) repo into this repo so the two copies stay in parity. The `predict-omen` subgraph here was several changes behind; this brings it fully up to date. Addresses two issues plus the backlog of unported omen work (Brier score, daily settled fields, agent-ID filtering, dead-code cleanup).

## predict-omen

- **Agent ID filtering**: two-step `RegisterInstance` + `TraderService` gate; `PREDICT_AGENT_IDS = [14, 25]` (both Valory trader agent IDs on Gnosis, per https://olas.network/data — filtering on a single ID would silently drop one cohort).
- **`answerFinalizedTimestamp`** on `FixedProductMarketMakerCreation` (#114): set to `answer ts + question.timeout` on each `LogNewAnswer` (re-answers push it out), cleared on `LogNotifyOfArbitrationRequest` (finalization frozen during arbitration), set to `block.timestamp` on arbitrator `LogFinalize`. Consumers treat an answer as final only when `answerFinalizedTimestamp <= now` — fixes the ZD#919 case where a temporary `0xff…ff` invalid sentinel was cached as permanently invalid. The two newly-wired events do finalization tracking only; settlement for arbitrated answers still runs via the `LogNewAnswer` emitted in the same tx.
- **Brier score**: `brierSum`/`brierCount` on `DailyProfitStatistic`, `impliedProbability` on `Bet`, `brierSumApplied`/`brierCountApplied` on `MarketParticipant`, plus `dailyTradedSettled`/`dailyFeesSettled` and `category`/`language` parsing.
- **`ConditionPreparation`** keeps its Question-existence guard (event ordering is guaranteed) with a documenting comment.
- **Dead-code cleanup**: removed orphaned `Log*` entities and handlers.

## predict-polymarket

- **Head-to-head markets** (#128): `extractBinaryOutcomes` now accepts any 2-outcome market instead of requiring literal Yes/No labels, so head-to-head e-sports / match-winner markets (team-name outcomes) get a `Question`/`MarketMetadata` and settle into profit aggregates. >2-outcome lists are still rejected (binary-ness also enforced upstream via `outcomeSlotCount == 2`). Label parser hardened to truncate at the `, p3` clause before the `.` fallback so names like `Gen.G` / `St. Louis` survive intact.

## Test plan

- [x] `predict-omen`: 53 tests pass (`yarn codegen && yarn build && yarn test`)
- [x] `predict-polymarket`: 116 tests pass
- [x] Both subgraphs build clean

## Deploy note

Both subgraphs require a **full re-sync** to repair historical data — the relevant past events (`LogNewQuestion`/`QuestionInitialized`/`OrderFilled`) won't fire again, and the affected entities are immutable. For polymarket specifically, ROI/profit aggregates will shift retroactively as the previously-dropped head-to-head markets fold into the totals (this is the correction, not a regression — worth flagging to Polystrat / olas-website owners).

Closes #114
Closes #128


LOCKhart07 (approved):
Reviewed the full diff plus the surrounding code (ABI param order for `RegisterInstance`, the service-lifecycle ordering the agent-ID filter relies on, the `handleBuy`/`handleSell` guards, the Realitio ABI signatures for the two newly-wired events, and the upstream `outcomeSlotCount == 2` bridge invariant). Correct, internally consistent, and well-tested. No blocking issues — approving with two non-blocking notes.

**1. `impliedProbability` uses gross `investmentAmount` (fee-inclusive) → slightly upward-biased Brier.** `FixedProductMarketMakerMapping.ts` computes `investmentAmount * 1e18 / outcomeTokensBought`, but on Omen FPMM `investmentAmount` is gross — shares are bought with `investmentAmount − feeAmount`. So the implied probability (and the Brier score consuming it) is biased high by ~the market fee (~2%). For a calibration metric the net price `(investmentAmount − feeAmount) / outcomeTokensBought` is arguably more accurate. It's documented in the schema comment so it reads as deliberate — just confirming it's intentional and matches the upstream `autonolas-subgraph` source (since this is a parity sync).

**2. Doc nit (re-answer reversal comment in `realitio.ts`).** "*Pre-Brier participants (indexed before this deploy) default to zero, so the reversal is a no-op*" — graph-node doesn't default missing required fields; reading such an entity would error. This is only safe because the deploy note mandates a full re-sync (so no pre-Brier participants exist). Consider rewording to reference the mandatory re-sync so it isn't read as forward-deploy-safe.

Everything else verified clean: agent-ID filter ordering (`RegisterInstance` precedes `CreateMultisigWithAgents`), `answerFinalizedTimestamp` semantics (scheduled `ts+timeout`, `block.timestamp` on arbitrator `LogFinalize`, null during arbitration), polymarket binary-ness still enforced upstream, the `", p3"` truncation preserving `Gen.G`/`St. Louis`, and the PR description + re-sync deploy note matching the diff.

atepem (commented):
**Verdict: 👍 Approve with minor nits.** High-quality sync PR — logic is correct, the accounting invariants hold under re-answer chains, test coverage is excellent (66 new/changed cases across 5 files), and the CLAUDE.md docs are kept fully in sync. I verified the shared ABIs contain the newly-wired events with exactly matching signatures (`LogFinalize(indexed bytes32,indexed bytes32)`, `LogNotifyOfArbitrationRequest(indexed bytes32,indexed address)`, `RegisterInstance(indexed address,indexed uint256,indexed address,uint256)`), so codegen/build will succeed. Nothing below is blocking.

## What's correct

**predict-omen**
- **Agent-ID filtering** — clean two-step `RegisterInstance` → `TraderService` marker → `CreateMultisigWithAgents` gate. Indexing both `[14, 25]` avoids the silent single-cohort drop.
- **`answerFinalizedTimestamp`** — set on `LogNewAnswer`, frozen (null) on arbitration request, stamped on `LogFinalize`. The "settlement still runs via the same-tx `LogNewAnswer`" reasoning is sound; these two handlers only touch finality, never profit state. Good null guards on `question`/`id`/`fpmm`.
- **Brier score** — I traced the reversal carefully. The invariant (`participant.brierSumApplied` stores exactly what was credited, so a re-answer subtracts the exact amount and recomputes the full set against the new outcome) holds, including the same-day re-answer case where `oldStatId == newStatId` (the shared cache-object reference makes subtract-then-add net correctly). The A→B→A chained test confirms convergence. `dailyTradedSettled`/`dailyFeesSettled` follow the identical, already-proven full-cost reversal pattern. No overflow risk (BigInt); sells correctly excluded via the `bet.amount.gt(0)` check; zero-probability sentinel skip is consistent across both settlement paths.

**predict-polymarket**
- **Head-to-head markets** — dropping the `isYesNoPair` gate while still enforcing exactly-2-outcomes is the right call, and the `", p3"`-before-`"."` truncation ordering correctly preserves names like `Gen.G` / `St. Louis`.

## Minor observations (non-blocking)

1. **Empty category/language stored as `""` not `null`** — `FPMMDeterministicFactoryMapping.ts`. If a 4-field template has an empty category segment, `fields[2].trim()` stores `""`. The schema documents these as nullable, so consumers now have two "absent" sentinels (`null` and `""`). A `len > 0 ? trimmed : null` would be cleaner.

2. **`out1` parsing is asymmetric to `out2`** — `uma-mapping.ts`. Only `out2` got the `", p3"` hardening; `out1` still terminates at the first comma. Fine in practice (the inter-name comma reliably ends p1) and pre-existing, but if a `p1` competitor name ever contained a comma it'd truncate. Note only.

3. **`subgraph.yaml` Realitio `entities` list omits `MarketParticipant`** — `handleLogNewAnswer` calls `participant.save()` but the entity isn't declared. This list is advisory in graph-node (not runtime-enforced) and the omission is pre-existing, but since the list was edited here, adding `MarketParticipant` would make it accurate.

4. **`answerFinalizedTimestamp` stays null if `question.timeout` is null** — correctly guarded and tested. Fine under the documented full re-sync (every tracked `LogNewQuestion` sets `timeout`); just confirms consumers must treat null as "not yet final," which the schema comment already states.

## Deploy note
The new non-nullable schema fields (`Bet.impliedProbability`, `MarketParticipant.brier*Applied`, `DailyProfitStatistic.dailyTradedSettled`/`dailyFeesSettled`/`brier*`) plus the polymarket head-to-head correction both require the documented **full re-sync**, and polymarket ROI aggregates will shift retroactively as previously-dropped markets fold in. That's a correction, not a regression — worth flagging to Polystrat / olas-website owners before deploy, as the PR description already notes.

_Note: I did not run the Matchstick suites locally — CI runs codegen + test across the matrix for both targets, and the PR reports 53 omen / 116 polymarket passing._

atepem (dismissed):

kupermind (dismissed):

kupermind (approved):

atepem (approved):

PR #152, opened 2026-06-15
feat(governance): switch to GovernorOLAS V3

Tanya-atatakai:
## Summary

Switches the governance subgraph from the old governor to the new GovernorOLAS deployment on Ethereum mainnet, following the existing V1→V2 archival pattern.

- **Archived** old governor `GovernorOLASV2` (`0x8e84b5055492901988b831817e4ace5275a3b401`) by adding `endBlock: 25200699` (the V3 deployment block).
- **Added** new data source `GovernorOLASV3` (`0x060D0CBdDFb0498d610E2EF55C01516B5B1251E6`) starting at `startBlock: 25200699`.

The new governor is also a `GovernorOLAS` contract (verified on Etherscan), so it reuses the same ABI, entity set, and all 11 event handlers via the existing YAML anchors — no mapping/code changes needed.

| Contract | Address | Blocks | Status |
|----------|---------|--------|--------|
| GovernorOLAS V1 | `0x34C895…3dd5` | 15050305 → 17527057 | Archived |
| GovernorOLAS V2 | `0x8e84b5…b401` | 17527057 → 25200699 | Archived |
| GovernorOLAS V3 | `0x060D0C…51E6` | 25200699 → present | Active |

## Testing

- `yarn codegen` ✅
- `yarn build` ✅ (all 3 data sources compile)

### DIFF
--- a/subgraphs/governance/CLAUDE.md
+++ b/subgraphs/governance/CLAUDE.md
@@ -9,7 +9,7 @@ Tracks the OLAS token governance system on Ethereum mainnet. Indexes three Gover
 subgraphs/governance/
 ├── schema.graphql          # Entity definitions (11 entities)
 ├── subgraph.yaml           # Manifest with 3 GovernorOLAS data sources
-├── package.json            # graph-cli ^0.97.0, graph-ts ^0.38.0, matchstick-as 0.5.0
+├── package.json            # graph-cli 0.98.1, graph-ts 0.38.2, matchstick-as 0.6.0 (exact pins, no carets)
 ├── src/
 │   ├── governor-olas.ts    # All 11 event handlers
 │   └── utils.ts            # Historical quorum calculation utility
@@ -20,8 +20,8 @@ subgraphs/governance/
 | Contract | Address | Blocks | Status |
 |----------|---------|--------|--------|
 | GovernorOLAS V1 | `0x34C895f302D0b5cf52ec0Edd3945321EB0f83dd5` | 15050305 → 17527057 | Archived |
-| GovernorOLAS V2 | `0x8e84b5055492901988b831817e4ace5275a3b401` | 17527057 → 25200699 | Archived |
-| GovernorOLAS V3 | `0x060D0CBdDFb0498d610E2EF55C01516B5B1251E6` | 25200699 → present | Active |
+| GovernorOLAS V2 | `0x8e84b5055492901988b831817e4ace5275a3b401` | 17527057 → 25322326 | Archived |
+| GovernorOLAS V3 | `0x060D0CBdDFb0498d610E2EF55C01516B5B1251E6` | 25322326 → present | Active |
 
 All three contracts share the same ABI (`abis/GovernorOLAS.json`) and the same set of 11 event handlers. V1 and V2 have an `endBlock` set at their respective upgrade points.
 
@@ -141,7 +141,7 @@ yarn test                   # Run Matchstick tests (no tests currently)
 ## AI Summary
 
 ### Critical Points
-1. **Three contract versions**: V1 archived at block 17527057, V2 archived at block 25200699, V3 active from block 25200699. All use identical handlers and ABI.
+1. **Three contract versions**: V1 archived at block 17527057, V2 archived at block 25322326, V3 active from block 25322326. All use identical handlers and ABI.
 2. **Vote accumulation asymmetry**: `handleVoteCast` updates proposal vote tallies; `handleVoteCastWithParams` does not.
 3. **Lazy quorum**: Quorum is not set at proposal creation — it's backfilled on first lifecycle event (cancel/execute/queue) when `currentBlock > startBlock`.
 4. **Proposal entity is mutable**: Updated by VoteCast (tallies), ProposalCanceled/Executed/Queued (status flags), and quorum calculation.
--- a/subgraphs/governance/CLAUDE.md
+++ b/subgraphs/governance/CLAUDE.md
@@ -1,6 +1,6 @@
 # Governance Subgraph
 
-Tracks the OLAS token governance system on Ethereum mainnet. Indexes two GovernorOLAS contract versions, covering the full proposal lifecycle (creation, voting, queuing, execution, cancellation), governance parameter changes, and timelock management.
+Tracks the OLAS token governance system on Ethereum mainnet. Indexes three GovernorOLAS contract versions, covering the full proposal lifecycle (creation, voting, queuing, execution, cancellation), governance parameter changes, and timelock management.
 
 ## Architecture Overview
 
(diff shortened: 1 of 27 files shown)


kupermind:
## Review — analysis

I verified the on-chain facts behind this switch against mainnet (public RPC, head block ~25324139). The mechanical change is clean, but **the core premise doesn't hold on-chain, and merging as-is would silently stop indexing live governance.** Requesting changes.

### What checks out ✅
- **V3 is a real GovernorOLAS.** `name()` → `"Governor OLAS"`, `version()` → `"1"` — byte-identical to V2. Same ABI/handlers reuse is valid.
- **V3 address is correct & EIP-55 valid** (`0x060D0C…51E6`); the YAML anchor reuse (`*GovernorOLASEntities` / `*GovernorOLASEventHandlers`) is correct; `yarn build` compiling all 3 sources matches what I'd expect.
- **V3 deployment block is exactly 25200699.** `eth_getCode` is empty at 25200698 and non-empty at 25200699. So the `startBlock` itself is right — no missed-deploy gap.

### Blocking issue ❌ — V2 is still the active governor; V3 is dormant
`endBlock: 25200699` on V2 is derived from *when V3 was deployed*, but deployment ≠ governance migration. Querying both contracts' logs from 25200699 → head:

| Contract | Events after block 25200699 | Latest event |
|---|---|---|
| **V2** (being archived) | **11** — incl. 2× `ProposalCreated`, multiple `VoteCast`, `ProposalQueued`, `ProposalExecuted` | block **25322326** (~1.8k blocks below head) |
| **V3** (being activated) | **0** | — none ever |

Decoded V2 events after the proposed cutoff (topic0-verified):
- 25278969 `ProposalCreated`, 25289543 `ProposalCreated` — **new proposals created on V2 after V3 existed**
- 25294609 / 25301032 / 25314113 / 25321675 `VoteCast`
- 25314138 / 25322296 `ProposalQueued`, 25321119 / 25322326 `ProposalExecuted`

So governance is running a full create→vote→queue→execute lifecycle **on V2, right up to chain head**, while V3 has emitted nothing. Setting `endBlock: 25200699` would make the subgraph **drop every governance event from late-May 2026 onward** (≥11 already, and counting), freezing the governance UI — and the newly-added V3 source would index an empty contract.

CI can't catch this: the matrix only runs `codegen`/`build`, which validate YAML+ABI shape, not whether the cutover block matches reality.

### Recommendation
This PR looks **premature** rather than wrong-in-mechanics. V3 is deployed but not yet adopted. Two clean options:
1. **Hold the switch.** Keep V2 with **no `endBlock`** until governance actually migrates (first `ProposalCreated` on V3 / last activity on V2), then set V2's `endBlock` to the *real* handover block — the way V1→V2 used the actual handover (17527057), not a deploy block.
2. **Forward-prep only.** If you want V3 indexed now, add the V3 data source **without** truncating V2 (leave V2 open). V3 indexes nothing until it's used; V2 keeps capturing the live lifecycle. Add the `endBlock` to V2 in a follow-up once the cutover happens.

Either way, please **don't set `endBlock` on V2 at 25200699** — that's the one line that causes data loss. Happy to re-verify once the migration actually lands on-chain.

<sub>Method: `eth_getCode` boundary check for the deploy block; chunked `eth_getLogs` per address over (25200699, head], topic0 matched against the 11 GovernorOLAS event signatures. Reproducible against any mainnet RPC.</sub>

subgraphs/governance/CLAUDE.md
```diff
@@ -20,9 +20,10 @@ subgraphs/governance/
 | Contract | Address | Blocks | Status |
 |----------|---------|--------|--------|
 | GovernorOLAS V1 | `0x34C895f302D0b5cf52ec0Edd3945321EB0f83dd5` | 15050305 → 17527057 | Archived |
-| GovernorOLAS V2 | `0x8e84b5055492901988b831817e4ace5275a3b401` | 17527057 → present | Active |
+| GovernorOLAS V2 | `0x8e84b5055492901988b831817e4ace5275a3b401` | 17527057 → 25200699 | Archived |
+| GovernorOLAS V3 | `0x060D0CBdDFb0498d610E2EF55C01516B5B1251E6` | 25200699 → present | Active |
```
atepem: Doc will be inaccurate under either resolution of the timing issue @kupermind raised. Since V3 has emitted no events and V2 is still live, V3 is not `Active` / `→ present` yet, and V2 is not `Archived`. Whichever path the team picks, please keep this table coupled to it:
- **Hold the switch** → revert this table; V2 stays `Active`.
- **Forward-prep (leave V2 `endBlock` unset)** → V2 stays `Active`; label V3 something like "deployed, not yet adopted" rather than `→ present | Active`.

(Note: the YAML boundary itself is mechanically fine — V2 `endBlock` and V3 `startBlock` share block 25200699, but since both bounds are inclusive *and the addresses differ*, there's no gap or double-indexing. The only problem is the cutover timing already flagged.)

subgraphs/governance/CLAUDE.md
```diff
@@ -1,14 +1,14 @@
...
 
 ## Architecture Overview
 
 ### Directory Structure
 ```
 subgraphs/governance/
 ├── schema.graphql          # Entity definitions (11 entities)
-├── subgraph.yaml           # Manifest with 2 GovernorOLAS data sources
+├── subgraph.yaml           # Manifest with 3 GovernorOLAS data sources
 ├── package.json            # graph-cli ^0.97.0, graph-ts ^0.38.0, matchstick-as 0.5.0
```
atepem: Minor / pre-existing (not introduced by this PR, but cheap to fix while this file is already being edited): this line is stale. The actual `package.json` uses exact pins — `@graphprotocol/graph-cli` `0.98.1`, `@graphprotocol/graph-ts` `0.38.2`, `matchstick-as` `0.6.0` — matching the repo-wide convergence noted in the root CLAUDE.md. Suggest updating to `graph-cli 0.98.1, graph-ts 0.38.2, matchstick-as 0.6.0` (no carets).

Tanya-atatakai: Good catch — fixed. Updated to the exact repo-wide pins: `graph-cli 0.98.1, graph-ts 0.38.2, matchstick-as 0.6.0` (no carets).

subgraphs/governance/CLAUDE.md
Tanya-atatakai: Table is now coupled to the resolution. We took the **handover** path: the cutover block has been moved to the real V2→V3 governance handover (**25322326**), not the V3 deploy block. Updated the table to V2 `17527057 → 25322326 | Archived` and V3 `25322326 → present | Active`, and the AI-summary block numbers to match. So `Archived`/`Active` now reflect the actual on-chain handover.

Tanya-atatakai:
@kupermind thanks for the thorough on-chain verification — you nailed it. `25200699` was the V3 **deploy** block, not the governance handover, so cutting V2 there would have silently dropped the live V2 lifecycle (the 11+ events you decoded) and started V3 against an empty contract.

Fixed: V2 `endBlock` and V3 `startBlock` are now both set to **25322326**, the actual V2→V3 governance handover, rather than the deploy block. This keeps V2 indexed through the real cutover (no dropped governance events) and starts V3 at adoption.

Also synced `subgraphs/governance/CLAUDE.md` to match — V2 `17527057 → 25322326 | Archived`, V3 `25322326 → present | Active`, and the AI-summary block numbers — so the docs no longer reference the deploy block. PTAL.

atepem (dismissed):

kupermind (dismissed):

kupermind (approved):

atepem (approved):

PR #156, opened 2026-06-23
feat(babydegen-base): scaffold Basius subgraph on Base (Aerodrome)

kupermind:
## What

New subgraph **`babydegen-base`** for the **Basius** agent (Pearl) on **Base** — a port of `babydegen-optimism`, Aerodrome-only. Plus a repo-wide `undici` security bump (see Supply-chain note) and a Phase-2 daily-activity stub.

> This description was updated after review to match the code. Earlier drafts described a single "service-115 pin / agent id 9" design — that was a config mix-up, now corrected (see below).

## Agent identity — **agentId 115** (confirmed on-chain)

Indexing filters on `event.params.agentId == 115` and tracks **every** service registered under that agent, mirroring optimism's `OPTIMUS_AGENT_ID` filter (no single-service pin). Verified against Base `ServiceRegistryL2 0x3C1fF68f…`:
- agentId-115 services: **607, 610, 611, 612** (610 confirmed via `RegisterInstance(... agentId=115)`; 610/611/612 all carry agentId `[115]`).
- **`115` is the agent id, not a service id.** Service *id* 115 is an unrelated 2025 service on the generic agentId 9 — explicitly excluded by the filter.
- `startBlock = 47163056` (earliest agentId-115 registration, service 607).

## Scope / port

- **Aerodrome only** (Velodrome fork — `velo*` handlers + `abis/defi/Velodrome*` reused; protocol strings → `aerodrome-cl`/`aerodrome-v2`). Uniswap V3 + Balancer dropped.
- Chainlink primary pricing (Base feeds verified live). **OLAS dropped** (Basius holds none); **AERO priced** off the Aerodrome AERO/USDC volatile pool. Whitelisted stables ≈ $1 via USDC feed.
- All addresses confirmed (registry, Aerodrome v2 PoolFactory, Slipstream NFPM/factory, LiFi, Chainlink) — by Divya and/or on-chain.

## KPIs (already produced by the ported pipeline)

`DailyPopulationMetric.sma7dAPR` = APR-rel-USDC-MA7D; `sma7dEthAdjustedAPR` = APR-rel-ETH-MA7D; `AgentPortfolioSnapshot.roi` = per-day ROI.

## Phase 2 (stub)

`DailyActivityMetric { dayTimestamp, activeAgents, transactionCount }` + per-day dedup. `transactionCount` counts LiFi swaps — **provisional** pending the product definition of "transactions" (swaps vs Safe executions vs mech requests). DAA = distinct services active/day.

## Supply-chain note

Includes the `undici ^7.24.0 → ^7.28.0` bump (root + 13 subgraphs) that fixes 3 HIGH advisories. This was **folded in here by maintainer decision** (originally split out as #157) so this branch is self-sufficient; `.supply-chain/install-hooks.allowlist` re-checked — no drift (within-major bump, 0 hooks).

## Test plan

`yarn codegen` + `yarn build` + `yarn test` (13/13) + `lockfile-lint` + `audit:prod` (0 high/crit) — all green in CI.

Follow-up stacked in #158 (Aerodrome ABI-compat verification + LpSugar bootstrap fix).


DIvyaNautiyal07:
Hey @kupermind, here's our pass on the open questions and a correction on the four addresses we originally handed you. Sorry for the bad input, those were stale Optimism values that got mislabelled as Base on our end.

**The four addresses, corrected**

We re-verified each against on-chain bytecode. Verdict:

```
ServiceRegistry  0x48b6af7B…  → wrong. This is the Ethereum L1 ServiceRegistry.
                                Use 0x3C1fF68f5aa342D296d4DEe4Bb1cACCA912D95fE
                                (canonical L2 registry, what you already swapped in).

StakingToken     0x88996bbd…  → wrong, that's the Optimism address.
                                Use 0x2585e63df7BD9De8e058884D496658a030b5c6ce on Base.

ActivityChecker  0x7Fd1F4b7…  → wrong, that's the Optimism/Mode/Arbitrum address.
                                Use 0x87C9922A099467E5A80367553e7003349FE50106 on Base.

Multisend        0xbE5b0013…  → wrong, that's the Optimism MultiSend.
                                Use 0x998739BFdAAdde7C933B942a68053933098f9EDa on Base.
```

All four corrected addresses are deployed on Base only, zero bytes on Optimism / Mode / Ethereum, so they're unambiguously the Base contracts.

**Answers to the open questions**

1. **Canonical registry `0x3C1fF68f…`.** Confirmed, use it. That's what Basius itself runs against.

2. **Service-115 pin vs agent-id-9 filter.** Keep the service-115 pin. Agent id 9 is generic and other Base services could end up using it; we don't want this subgraph absorbing unrelated services. If a second Basius-equivalent service ships on Base we'll revisit, but for now single-service pin is the right call.

3. **Aerodrome v2 PoolFactory.** `0x420DD381b31aEf6683db6B902084cB0FFECe40Da` confirmed (7032 bytes of bytecode on Base, matches Aerodrome's published address). Use it.

4. **service-115 registration block.** Will follow up with the exact block. Basius is very fresh (the agent package only landed in our repo on 2026-06-22), so the registration is in the most recent slice of Base history. Easiest way to pin it: open `0x3C1fF68f…` on Basescan, filter `CreateService` for serviceId=115, take the block of that log.

5. **LiFi Diamond on Base.** `0x1231DEB6f5749EF6cE6943a275A1D3E7486F4EaE` confirmed. LiFi's CREATE2 deployment, same address on every EVM chain. Bytecode present on Base.

6. **OLAS / AERO pricing + whitelisted stables.** For the stables (BOLD, msUSD, frxUSD, eUSD, axlUSDC), $1 via the USDC feed is fine. They all peg to USD and Basius doesn't hold meaningful balances, so the fidelity loss is negligible. For OLAS and AERO we'll send specific Aerodrome pool addresses (deepest WETH or USDC pair) in a follow-up. Both can stay at $0 in the first deploy and we'll backfill in a follow-up PR. AERO matters more than OLAS since it's the CL gauge reward, so prioritise that one.

7. **Block-handler cadence.** Keep `every: 1800`. Snapshots only fire on UTC-midnight crossings, so dropping to 900 doubles the trigger count for the same number of actual snapshots. The ~1h cadence is plenty.

8. **Phase 2 transactions / DAA.** @Tatiana could you confirm how this should work? Tagging you to define "transactions per day" (mech requests vs swaps vs something else) and how DAA should be computed for babydegen so we can scope Phase 2 properly.

Everything else you listed (tokens, whitelisted stables, the full Aerodrome surface, placeholder v2 PoolFactory and LiFi) checks out on Base. So aside from the four corrections above, the rest of your config is safe to keep as-is.

subgraphs/babydegen/babydegen-base/src/serviceRegistry.ts
```diff
@@ -0,0 +1,175 @@
...
+  return false
+}
+
+export function handleRegisterInstance(event: RegisterInstance): void {
+  // PIN to the single Basius service (id 115). Basius registers under canonical
+  // agent id 9, but agent 9 may be shared by other Base services, so we pin by
+  // service id rather than agent id.
+  // TODO(reviewers): should this be generalised to an agent-id filter instead of
+  // a single-service pin? (see IMPLEMENTATION-PLAN.md open questions)
+  if (!event.params.serviceId.equals(BASIUS_SERVICE_ID)) {
```
Tanya-atatakai: **Blocker — wrong field.** Filter on `agentId`, not `serviceId`:
```ts
if (!event.params.agentId.equals(BASIUS_AGENT_ID)) { return }
```
Basius = `agentId == 115` and **all** such services should be tracked. Pinning to a single `serviceId` tracks only one service (and matches on the wrong attribute). This is exactly what babydegen-optimism does at the equivalent line.

subgraphs/babydegen/babydegen-base/src/serviceRegistry.ts
```diff
@@ -0,0 +1,175 @@
...
+    }
+  }
+}
+
+export function handleCreateMultisigWithAgents(event: CreateMultisigWithAgents): void {
+  let serviceId = event.params.serviceId
+  let multisig = event.params.multisig
+
+  // PIN to the single Basius service (id 115).
+  if (!serviceId.equals(BASIUS_SERVICE_ID)) {
```
Tanya-atatakai: Remove this `serviceId == 115` pin. Once `handleRegisterInstance` filters by agentId, the existing `if (registration == null) return` below already drops non-Basius services — this is how optimism does it. As written, this hard-caps the subgraph to a single service even after the RegisterInstance fix.

subgraphs/babydegen/babydegen-base/src/constants.ts
```diff
@@ -0,0 +1,93 @@
...
+export const VELO_NFT_MANAGER = Address.fromString("0xe1f8cd9AC4e4A65F54f38a5CdAfCA44f6dD68b53") // Aerodrome Slipstream NFPM (CL positions)
+export const VELO_MANAGER = Address.fromString("0xe1f8cd9AC4e4A65F54f38a5CdAfCA44f6dD68b53")     // alias of the Slipstream NFPM
+export const VELO_FACTORY = Address.fromString("0xf8f2eB4940CFE7d13603DDDD87f123820Fc061Ef")     // Aerodrome Slipstream (CL) factory
+export const VELO_V2_FACTORY = Address.fromString("0x420DD381b31aEf6683db6B902084cB0FFECe40Da")  // TODO(divya): VERIFY Aerodrome v2 PoolFactory (not in Divya's list)
+export const VELO_V2_SUGAR = Address.fromString("0x69dD9db6d8f8E7d83887A704f447b1a584b599A1")     // Aerodrome LpSugar v3 (pool discovery / bootstrap)
+
+// ---- Olas service configuration (Base) --------------------------------------
+// Canonical ServiceRegistryL2 on Base = 0x3C1fF68f5aa342D296d4DEe4Bb1cACCA912D95fE (set in subgraph.yaml).
+// Basius = service 115, canonical agent ID 9, service safe (multisig) 0x9eb5faed6e6983fedc4206af1b58a17fabe9a0d9.
+export const BASIUS_SERVICE_ID = BigInt.fromI32(115) // we PIN indexing to this single service (see serviceRegistry.ts)
```
Tanya-atatakai: These should become a single `BASIUS_AGENT_ID = BigInt.fromI32(115)` used for filtering; `BASIUS_SERVICE_ID` and the `agent ID 9` framing can go. The comment "we pin by service id" encodes the bug.

subgraphs/babydegen/babydegen-base/tests/mapping.test.ts
```diff
@@ -0,0 +1,379 @@
...
+    )
+    assert.fieldEquals(
+      "ServiceRegistration",
+      id.toHexString(),
+      "registrationTxHash",
+      TX_HASH.toHexString()
+    )
+  })
+
+  test("registers regardless of agent ID (pinned by service id, not agent id)", () => {
```
Tanya-atatakai: These two tests lock in the wrong behavior ("pinned by service id, not agent id" / "ignores non-pinned service IDs"). After the fix they should assert: non-Basius **agentId** → 0 registrations, and two different serviceIds both with agentId 115 → 2 registrations (proving multi-service tracking).

subgraphs/babydegen/babydegen-base/subgraph.yaml
```diff
@@ -0,0 +1,464 @@
...
+# TODO(divya): all `startBlock` values are placeholders (17310019) — set them to
+# the block where Basius service 115 was registered / first funded on Base.
+dataSources:
+  - kind: ethereum
+    name: ServiceRegistryL2
+    network: base
+    source:
+      address: "0x3C1fF68f5aa342D296d4DEe4Bb1cACCA912D95fE" # canonical Olas ServiceRegistryL2 on Base
+      abi: ServiceRegistryL2
+      startBlock: 17310019 # TODO: set to service-115 registration block
```
Tanya-atatakai: `startBlock` placeholder. With an agentId filter, set this to the **earliest agentId-115 service registration block** (not "service-115 registration block"). A too-late startBlock silently drops early services. Applies to all the duplicated `17310019` values below too.

Tanya-atatakai (changes requested):
## Service scoping uses the wrong identifier (blocker)

Thanks for the clean port — the velo→aerodrome naming rationale, the Uniswap/Balancer drops, and the honest `TODO` marking are all great. One blocking correctness issue plus a few secondary notes.

**Blocker:** Basius services must be identified by **`agentId == 115`** (tracking *all* such services), not by a single pinned `serviceId == 115`. This PR pins to service id 115 in both `serviceRegistry.ts` handlers and treats agent id as `9`/"context only" — which is backwards.

Why it matters:
- `RegisterInstance(address, uint256 serviceId, address, uint256 agentId)` exposes `event.params.agentId`. The sibling **babydegen-optimism** already does the right thing: `if (!event.params.agentId.equals(OPTIMUS_AGENT_ID)) return`, and lets `CreateMultisigWithAgents` filter naturally via `ServiceRegistration.load() == null` — no per-service pin.
- A single-`serviceId` pin tracks **only one** service and drops every other agentId-115 service. It also matches on the wrong attribute entirely.
- The rest of the code is already multi-service capable (`config.ts:getServiceByAgent`, multisig-keyed snapshots). The pin is the *only* thing breaking that contract.

Fix (mirror optimism): set `BASIUS_AGENT_ID = 115`, filter `handleRegisterInstance` on `event.params.agentId`, drop the `serviceId` pin from `handleCreateMultisigWithAgents` (keep the existing `registration == null` guard), and flip the two filter tests to assert agentId-based behavior.

**Secondary:** all `startBlock`s are placeholder `17310019`; with an agentId filter this should be the earliest agentId-115 registration block. Placeholder addresses (`VELO_V2_FACTORY`, Chainlink feeds, LiFi Diamond) and unpriced OLAS/AERO are correctly TODO-marked but will distort valuations until resolved. Stale "service 115 / agent 9 / Optimus" framing in `CLAUDE.md`/`constants.ts`/`serviceRegistry.ts:90` should be updated alongside the fix.

kupermind:
Pushed `9cda505` addressing the review. Thanks @Tanya-atatakai — your blocker was exactly right, and it turns out it traces back to a **service-id vs agent-id mix-up in the original config**, so this clears up @DIvyaNautiyal07's answer #2 too.

## On-chain proof (why agentId 115, and why "service 115" was a red herring)

I verified against the canonical registry `0x3C1fF68f…` on Base:

- **Multiple fresh DEPLOYED services run `agentId == 115`** — services **607, 610, 611, 612**, each with a *different* multisig. The earliest (#607) was created at block **47163056 (2026-06-10)** — consistent with Basius being new. These are the real Basius services.
- **"Service 115"** (the thing we were pinning to) has `agentIds: [9]` and was created **2025-03-28** — over a year before the Basius agent package existed. It's an unrelated old service on the *generic* agent 9. The date alone rules it out as Basius.

So `115` is the **agent id**, not a service id, and there are already several Basius services — exactly Tanya's point that a single-service pin is both the wrong attribute and too narrow.

## Applied (commit `9cda505`)

**Blocker fix (mirrors babydegen-optimism):**
- `handleRegisterInstance` filters `event.params.agentId == BASIUS_AGENT_ID` (115); `handleCreateMultisigWithAgents` drops non-Basius services via the existing `registration == null` guard (no pin).
- Single `BASIUS_AGENT_ID = BigInt.fromI32(115)`; removed the `serviceId`/`agent 9`/"pin" framing and stale Optimus comments.
- Tests flipped: non-Basius `agentId` → 0 registrations; two serviceIds both on agentId 115 → 2 registrations (multi-service).

**Divya's confirmations + my on-chain checks (TODOs cleared):**
- `startBlock` → **47163056** everywhere (earliest agentId-115 service).
- v2 PoolFactory `0x420DD381…`, LiFi `0x1231DEB6…`, Chainlink ETH/USD `0x71041ddd…` + USDC/USD `0x7e860098…` — all confirmed (the two feeds verified live on Base, 8-dec). Cadence `every: 1800` kept. Stables price $1 via the USDC feed.
- Removed the speculative **USDbC** (Basius funds in USDC native).
- Corrected Base StakingToken/ActivityChecker/Multisend documented as reference-only (babydegen doesn't index them).

`codegen` + `build` + **10/10 tests** green.

## Still open (tracked in IMPLEMENTATION-PLAN.md)

- **@DIvyaNautiyal07** — Aerodrome **OLAS/AERO price pools** (deepest WETH or USDC pair). Both stay $0 in this first cut; **AERO** is the priority since it's the CL-gauge reward token (reward USD reads 0 until then).
- **@Tatiana** — Phase 2 definitions: "transactions per day" (mech requests vs swaps) and how DAA is computed for babydegen.

> Note on CI: the red check here is the repo-wide `undici` advisory (graph-cli→undici), fixed in #157 — inherited from `main`, not this PR. Once #157 lands I'll rebase and it goes green. @Tanya-atatakai could you take a look at #157 too? It needs 2 approvals to merge.

@Tanya-atatakai — could you re-review when you have a moment?

DIvyaNautiyal07:
For #6: AERO uses Aerodrome V2 USDC/AERO volatile pool `0x6cdcb1c4a4d1c3c6d054b27ac5b77e89eafb971d` (~$26M TVL, deepest by an order of magnitude — `aerodrome_v2` sourceType, pairToken=USDC).

For OLAS: we couldn't find a usable Aerodrome pool on Base. The only OLAS pair that exists at the factory is OLAS/USDC v2 volatile (`0x0ea2b1d257f40c3dd0557b44aff96f5d590cbba2`), with reserves of ~0.76 OLAS / 0.01 USDC — effectively empty. No OLAS/WETH or OLAS/AERO pools exist on Aerodrome at all.

kupermind:
## Phase 2 (explorer daily metrics) — proposed definitions, please confirm

I've landed a **stub** for the agent-explorer heatmap (per-day `DailyActivityMetric` with `activeAgents` + `transactionCount`) so the skeleton exists, but two definitions are product calls. Tagging @Tatiana / @Presh / @DIvyaNautiyal07 to pin them down — a quick "this one" unblocks the final version.

### 1. What counts as a "transaction" per day?

This is the only part that's genuinely ambiguous, and it changes both the number and the build:

| Option | Source | Status |
|---|---|---|
| **A. Swaps** | `SwapTransaction` (LiFi) — already indexed | **stubbed now** (current `transactionCount`) |
| **B. Safe executions** | `ExecutionSuccess` on the agent Safe — already indexed | trivial to switch to |
| **C. Mech requests** | Base mech marketplace — **not indexed**; needs a new data source keyed to the Basius safes | larger build |

The Slack thread had this exact tension — @Tatiana noted babydegen historically doesn't use mechs (so "transactions" may not mean mech requests), while @DIvyaNautiyal07 said it now relies on mech requests. **Which row should the heatmap show?** (A is live today; C is the only one needing new indexing.)

### 2. How is DAA (daily active agents) defined?

Proposed: **a Basius service counts as "active" on a UTC day if it had ≥1 tracked on-chain action that day.** The stub currently keys "active" off swaps; I can broaden "action" to include funding, balance changes, and position open/close if you'd prefer a looser definition. For a (currently) small set of Basius services this is a per-day distinct-service count.

### What's stubbed vs pending

- ✅ `DailyActivityMetric { dayTimestamp, activeAgents, transactionCount }` + per-day dedup — **live** (transactionCount = swaps, provisional).
- ⏳ Final `transactionCount` semantics — your pick from the table (defaults to swaps).
- ⏳ Mech-requests variant (option C) — I'll add the data source only if that's the intended metric.

Happy to flip to safe-executions or wire the mech-marketplace source as soon as you confirm.

kupermind:
### Status + who's needed 🙏

Two commits since the last round — `9cda505` (agentId-115 fix) and `5bedb1a` (OLAS dropped, AERO priced, Phase 2 stub). CI is green.

**@Tanya-atatakai** — your blocker is addressed: filtering is now by `agentId == 115` (tracking all such services, mirroring optimism), the `serviceId` pin is gone, and the two filter tests are flipped to assert multi-service behavior. On-chain confirmed it: agentId-115 runs on services 607/610/611/612; "service 115" was an unrelated 2025 service on the generic agent 9. **Re-review requested** whenever you have a moment.

**@DIvyaNautiyal07** — all your config answers are in, thank you. The OLAS/AERO item is resolved without waiting on you: OLAS is **dropped** (Basius holds none), and AERO is now priced off the Aerodrome AERO/USDC volatile pool `0x6cdcb1c4…`. Nothing pending on your side unless you'd prefer a different AERO pool. Your input on the Phase 2 "transactions" question below is still welcome.

**Presh** (couldn't find your GitHub handle — @kupermind can you tag/forward?) — you own the explorer-heatmap metrics. The Phase 2 definition question is in [this comment](https://github.com/valory-xyz/autonolas-subgraph-studio/pull/156#issuecomment-4779905141): what should "transactions per day" count — **swaps** (live now), **Safe executions**, or **mech requests** (needs a new data source)? And how should **DAA** be defined? A quick pick unblocks the final Phase 2.

> Merge needs 2 approvals (branch protection). I can't self-approve — @kupermind, please add a second reviewer.

Tanya-atatakai (dismissed):
Re-reviewed — blocker resolved, approving. ✅

**Blocker (agentId vs serviceId) — fixed and verified on-chain.**
- `handleRegisterInstance` now filters `event.params.agentId == BASIUS_AGENT_ID` (115); the `serviceId` pin is gone from `handleCreateMultisigWithAgents` (the `registration == null` guard handles non-Basius services, mirroring babydegen-optimism).
- Collapsed to a single `BASIUS_AGENT_ID = BigInt.fromI32(115)`; serviceId/agent-9/"pin" framing removed across `constants.ts`, `serviceRegistry.ts`, and `CLAUDE.md`.
- Tests flipped exactly as requested: non-Basius agentId → 0 registrations; two serviceIds both on agentId 115 → 2 (multi-service tracking proven).
- `startBlock` → 47163056 (earliest agentId-115 service) on all data sources.

Thanks for the thorough on-chain confirmation (agentId 115 → services 607/610/611/612; "service 115" was the unrelated 2025 agent-9 service). Good catch folding the standalone undici #157 in here — the `^7.28.0` bump across root + all subgraphs looks clean and all 34 checks pass.

Secondary items (placeholder pricing pools for AERO follow-up, Phase 2 transactions/DAA definitions) are tracked and don't block this scaffold. LGTM.

kupermind:
Thanks @Tanya-atatakai 🙏 — one small correction to your note: **AERO is already priced in this PR** (commit `5bedb1a`), not a follow-up — wired off the Aerodrome AERO/USDC volatile pool `0x6cdcb1c4…` via the `velodrome_v2` adapter (the VELO analog). The only genuine follow-ups left are the **Phase 2 "transactions" definition** (Tatiana/Presh) and broadening DAA if desired.

Branch protection needs **2 approvals** and this is at 1/2 — @kupermind, could you add a second reviewer so it can merge?

atepem (commented):
## Review summary

A near-mechanical port of `babydegen-optimism` to Base/Aerodrome (Velodrome fork, ABIs reused), plus a Phase-2 daily-activity stub and a repo-wide `undici` bump. The ported pipeline (schema, ROI/APR/snapshot, velo→aero renames, dropped UniV3/Balancer) is clean and consistent with the repo's multi-network patterns. The concerns below are in the agent-identity config the whole subgraph hinges on, and in PR hygiene.

### 🔴 Blocking — the agent identity is internally contradicted

Everything indexed depends on one line: `serviceRegistry.ts` filters `event.params.agentId.equals(BASIUS_AGENT_ID)` with `BASIUS_AGENT_ID = 115`. The PR's own artifacts disagree about what Basius *is*:

- **Code + `CLAUDE.md` + `IMPLEMENTATION-PLAN.md`:** "Basius = **agentId 115**, services 607/610/611/612," and explicitly warn that *service* 115 is an unrelated 2025 service on agentId **[9]**.
- **`README.md` + this PR description:** "pinned to Olas **service ID 115** … **canonical agent id 9**, safe `0x9eb5fa…`."

These describe two different on-chain entities. If the true Basius agentId is **9** rather than 115, the `agentId == 115` filter matches **nothing** and the subgraph silently indexes zero services — it would still build and pass tests (tests assert the filter, not the chain). Please confirm against Base `ServiceRegistryL2` that services 607/610/611/612 actually emit `RegisterInstance` with `agentId == 115` before merge. Highest-risk item, currently unverifiable from inside the repo.

### 🟠 README.md and the PR description are stale

The code moved from the old "single service-115 pin / TODO placeholders" design to "agentId-115 multi-service, addresses confirmed, Phase-2 stub landed," but `README.md` and the PR body still describe the service-115 pin, "canonical agent id 9," and unresolved TODO placeholders. Update both to match `CLAUDE.md`/`IMPLEMENTATION-PLAN.md` — or, if the README is the correct one, see the blocking item above.

### 🟡 Scope creep + supply-chain process

- The `undici ^7.24.0 → ^7.28.0` bump touches **all 14 `package.json` + `yarn.lock` files** and is unrelated to scaffolding a new subgraph. Per the repo's dependabot cadence it belongs in its own PR; bundling couples two independent risk surfaces and bloats the feature diff.
- Root `CLAUDE.md` requires refreshing `.supply-chain/install-hooks.allowlist` after any dep change. This PR touches no `.supply-chain/` files. A within-major undici bump almost certainly adds no install hooks, but the documented process is to re-run `yarn audit:install-hooks:update` and commit the result (or confirm no drift).

### 🟢 Notes (non-blocking)

- **Phase-2 stub** (`src/dailyActivity.ts`) is correct for what it claims — immutable `<day>-<safe>` dedup marker, DAA once/day, `transactionCount` per swap — and honestly labeled PROVISIONAL. Be aware the `transactionCount` semantics ship now and will need redefinition/backfill once product decides swaps-vs-Safe-exec-vs-mech.
- The USDC/WETH `Transfer` data sources index *all* transfers of very high-volume Base tokens from block 47M and filter in-handler — heavy, but inherited from the optimism design, not introduced here.

### Verdict

The mechanical port is merge-quality. Recommend **blocking on the agentId-115-vs-9 confirmation** (correctness-critical and currently self-contradictory), requiring the **README/PR-body sync**, and **splitting out the undici bump**. The "keep pin or generalize" scoping question the PR raises is effectively already answered in code (agentId filter) — but only if 115 is the right id.

<sub>🤖 Review by Claude Code</sub>

atepem (dismissed):

kupermind:
Thanks @atepem — thorough review. Addressed all of it; pushed `ba2397c` and updated the PR body.

### 🔴 agentId-115 vs 9 — confirmed on-chain, was a stale-docs contradiction (not a code bug)

The code was right; the README/PR-body were stale. Verified against Base `ServiceRegistryL2 0x3C1fF68f…`:

- **Service 610 emits `RegisterInstance(operator, 610, agentInstance, agentId=115)`** (block 47552986) — the direct event-level confirmation you asked for.
- `getService` shows services **610/611/612 all carry agentId `[115]`** (state DEPLOYED); 607 is registered but not yet deployed (state 1).
- "Service id 115" is the unrelated **2025** service on the generic **agentId 9** — exactly what the code's comments warn about. The `agentId == 115` filter matches the real Basius services, not nothing.

So the filter is correct and the "two different entities" were the README/PR text lagging the code.

### 🟠 README + PR body — synced

Both updated to the agentId-115 / multi-service reality (services 607/610/611/612, startBlock 47163056, no "service-115 pin / agent id 9 / single safe"). `ba2397c` for the README; PR description rewritten above.

### 🟡 Supply-chain

- **undici scope:** agreed it's logically separate — it was split out as #157 first, then **folded in here by maintainer decision** so this branch is self-sufficient (the root-tree audit gate fails on every branch until the bump lands). Noted explicitly in the PR body.
- **install-hooks allowlist:** good catch on the process. Re-ran `node scripts/audit-install-hooks.mjs --update` → **0 entries, no drift** in `.supply-chain/` (a within-major undici bump adds no install hooks, as you expected). `install-hooks: OK`.

### 🟢 Non-blocking — acknowledged

- Phase-2 `transactionCount` ships provisional (LiFi swaps) and is honestly labelled; redefinition/backfill tracked for the swaps-vs-Safe-exec-vs-mech decision (still need that from you/@Tatiana).
- The high-volume USDC/WETH `Transfer` indexing is inherited from the optimism design (filtered in-handler), not introduced here — noted as a known perf characteristic.

Separately, #158 (stacked) verifies the rest of the Aerodrome ABI reuse against the deployed contracts and fixes the one real mismatch (LpSugar `all`→`forSwaps`).

DIvyaNautiyal07 (dismissed):

kupermind:
### Open items — consolidated (everything else is resolved)

Thanks all — most of the config is now confirmed (agentId 115 verified on-chain, addresses, startBlock, AERO pricing). Two product calls remain, plus one deferred verification.

**✅ Confirmed from @DIvyaNautiyal07's pricing reply** (no action needed, just closing the loop):
- **AERO** is wired to the exact pool you gave — Aerodrome v2 **USDC/AERO volatile `0x6cdcb1c4…`**, pairToken USDC (~$26M TVL). (Internal adapter label is `velodrome_v2`, reused since Aerodrome is a Velodrome fork — same `getReserves` interface.)
- **OLAS dropped** — matches your finding that no usable Aerodrome OLAS pool exists (the only OLAS/USDC pair is ~empty). It's removed from the token set rather than priced at $0.

**❓ For @Tatiana / @atepem — Phase 2 (the one real open decision):**
1. **What counts as a "transaction per day"** on the explorer heatmap?
   - **Swaps** — live now (LiFi swaps), no extra work.
   - **Safe executions** (`ExecutionSuccess`) — cheap switch, already captured.
   - **Mech requests** — the only option needing a **new** Base mech-marketplace data source.
2. **DAA ("daily active agents")** — currently "a service that swapped that day." Should "active" broaden to *any* tracked activity (funding / balance change / position open-close)?

**⏳ Deferred (no one to ask — needs real data):** the Aerodrome CL-gauge `earned()` reward read is the same Slipstream-fork ABI that matched everywhere else and is `try_`-wrapped (degrades to 0 if wrong), but AERO reward valuation can't be confirmed end-to-end until a Basius service opens a *staked* CL position. The live services are brand-new (~$1.5 USDC each, no positions yet); I'll recheck once one exists.

Nothing above blocks the scaffold — #156 is functionally complete and CI-green. The Phase-2 transactions definition just determines whether the provisional swaps-based `transactionCount` stays or gets redirected (and whether we add the mech data source).

Tanya-atatakai:
Transactions-per-day and DAAs are already computed in the **service-registry** subgraph, so Phase 2 needs no new data source here.

- `DailyAgentPerformance` (`day-{dayTimestamp}-agent-{agentId}`) tracks `txCount` per agent per day — and that `txCount` is driven by Safe `ExecutionSuccess` / `ExecutionFromModuleSuccess`, which is exactly the "transaction per day" definition we want for the heatmap.
- `DailyUniqueAgents` (`day-{dayTimestamp}`) gives the deduped daily-active-agents count.

So nothing to build on the registry side — just need @atepem to confirm it's wired up to the heatmap.

kupermind:
Great catch @Tanya-atatakai — verified and acted on it. `service-registry` indeed already computes both, and it covers Base:

- `DailyAgentPerformance` (`day-{ts}-agent-{agentId}`, `txCount`) is incremented in `handleExecutionSuccess` / `handleExecutionFromModuleSuccess` — Safe-execution-driven, per `agentId`, so Basius (agentId 115) is already covered.
- `DailyUniqueAgents` (`day-{ts}`) for DAA.

So the babydegen-base Phase-2 stub was both **redundant** and **wrong-signal** (it counted LiFi swaps, not Safe executions). Removed it in `0ac1c0f` — dropped `DailyActivityMetric`/`DailyAgentActivity`, `src/dailyActivity.ts`, the LiFi hook, and the 3 stub tests. ROI/day still comes from this subgraph (`AgentPortfolioSnapshot.roi`); transactions + DAA come from `service-registry`. Docs updated to point there.

That closes the Phase-2 question on the subgraph side — the only remaining bit is consumer-side: @atepem wiring the heatmap to `service-registry`'s `DailyAgentPerformance.txCount` + `DailyUniqueAgents` for agentId 115. codegen + build + 10 tests green.

atepem:
I can wire up once its live :)

Tanya-atatakai (approved):

DIvyaNautiyal07 (approved):

PR #159, opened 2026-06-24
fix(babydegen-base): index AERO gauge rewards + value staked V2 positions (＋ CL verification plan)

kupermind:
## What & why

Started as a docs PR for the **CL** AERO-reward path; @DIvyaNautiyal07's on-chain verification (thank you!) turned it into a real fix. Her check used a live Basius **V2-gauge** stake, which confirmed most assumptions **and** surfaced a correctness gap in the V2 path. This PR now does three things:

1. **Fixes staked V2 positions vanishing from the portfolio** (the gap Divya's example exposed).
2. Adds a **silent-zero guard** on the CL reward path.
3. Documents the whole AERO-reward picture + the runbook to close out the one remaining CL gap (`AERO-REWARDS-PLAN.md`).

## 1. The V2 fix (main change)

V2 position value was read from `pool.balanceOf(safe)`. Staking LP into an Aerodrome V2 gauge **moves the LP out of the safe** → `balanceOf(safe) == 0` → the position was marked `isActive=false` and zeroed. So a staked, earning position **disappeared** from the portfolio. Inherited verbatim from `babydegen-optimism` (likely affects it too — follow-up).

`refreshVeloV2Position` now:
- **Resolves the gauge** via the Aerodrome **Voter** (`gauges(pool)`, `VELO_VOTER 0x16613524…`), cached on `rewardsContract`.
- Counts **staked LP**: `effectiveBalance = pool.balanceOf(safe) + gauge.balanceOf(safe)` (`pool.totalSupply` already includes staked LP, so share math is unchanged).
- Adds **rewards**: `gauge.earned(safe)` (AERO, priced off AERO/USDC) → `claimableReward/USD` and `usdCurrentWithRewards` (replaces the old `// TODO: Calculate V2 fees later`).

New ABIs: `abis/defi/VeloVoter.json`, `abis/defi/VeloV2Gauge.json`, wired into the `VeloV2Pool` template.

### Verified on-chain (Divya's live position, gauge `0x793f…`, block 47702491)
`Voter.gauges(pool)` → the gauge ✅ · `gauge.stakingToken()` == pool ✅ · `gauge.balanceOf(safe)` = `7157236262142` (== the agent's `current_liquidity`) ✅ · `gauge.rewardToken()` = AERO ✅ · `pool.balanceOf(safe)` = 0 (confirms the bug) ✅

## 2. CL silent-zero guard

CL gauges use a **different selector** (`earned(address,uint256)`) on a different gauge ABI than the V2 `earned(address)` Divya exercised. We can't claim CL is verified until a Basius position lands in a Slipstream pool, so `refreshVeloCLPosition` now emits a `log.warning` if `earned()` reverts on an active CL position — turning a silent 0 into a visible signal at the first real CL stake.

## 3. Verification status (`AERO-REWARDS-PLAN.md`)

- ✅ **Settled by Divya:** depositor == service safe (universal to V2+CL), `rewardToken` = AERO, `earned(safe)` resolves non-zero, deposited LP matches our record.
- ⏳ **Only strict gap left:** the CL `earned(address,uint256)` selector — event-gated on a real Slipstream stake; the `log.warning` is the trigger.

## Tests
`yarn codegen` + `build` + `test` green — **21/21** (＋2): a staked position (`pool.balanceOf==0`, gauge holds the LP) stays active with `liquidity` = staked amount; `gauge.earned` folds into `usdCurrentWithRewards` (~$2.60 for 2 AERO @ $1.30).

## Follow-ups (not here)
- Mirror the V2 staked-position fix to `babydegen-optimism` (same code).
- CL end-to-end confirmation once a Basius Slipstream position is staked.

### DIFF
--- a/abis/defi/VeloV2Gauge.json
+++ b/abis/defi/VeloV2Gauge.json
@@ -1,0 +1,30 @@
+[
+  {
+    "type": "function",
+    "name": "balanceOf",
+    "stateMutability": "view",
+    "inputs": [{ "name": "account", "type": "address" }],
+    "outputs": [{ "name": "", "type": "uint256" }]
+  },
+  {
+    "type": "function",
+    "name": "earned",
+    "stateMutability": "view",
+    "inputs": [{ "name": "account", "type": "address" }],
+    "outputs": [{ "name": "", "type": "uint256" }]
+  },
+  {
+    "type": "function",
+    "name": "rewardToken",
+    "stateMutability": "view",
+    "inputs": [],
+    "outputs": [{ "name": "", "type": "address" }]
+  },
+  {
+    "type": "function",
+    "name": "stakingToken",
+    "stateMutability": "view",
+    "inputs": [],
+    "outputs": [{ "name": "", "type": "address" }]
+  }
+]
--- a/abis/defi/VeloVoter.json
+++ b/abis/defi/VeloVoter.json
@@ -1,0 +1,9 @@
+[
+  {
+    "type": "function",
+    "name": "gauges",
+    "stateMutability": "view",
+    "inputs": [{ "name": "pool", "type": "address" }],
+    "outputs": [{ "name": "", "type": "address" }]
+  }
+]
--- a/subgraphs/babydegen/babydegen-base/AERO-REWARDS-PLAN.md
+++ b/subgraphs/babydegen/babydegen-base/AERO-REWARDS-PLAN.md
@@ -1,0 +1,147 @@
+# AERO gauge-reward indexing — verification & implementation plan
+
+**Status:** code path is **built and shipped** (in `#156`), but **unverified end-to-end**
+because no Basius CL position has been staked in an Aerodrome gauge on Base yet. This doc
+captures exactly what the path does, which assumptions are still unproven, and the concrete
+steps to confirm (and, if needed, fix) it once the first real staked position exists.
+
+This is deliberately a docs-first PR: there is **no Basius gauge state on-chain to test
+against today**, so we describe the plan and add self-reporting instrumentation rather than
+guess at fixes.
+
+---
+
+## 1. What the path does today
+
+For every **active Aerodrome Slipstream (CL)** position, `refreshVeloCLPosition`
+(`src/veloCLShared.ts`) reads unclaimed staking rewards from the position's gauge and folds
+them into the portfolio value:
+
+```ts
+// src/veloCLShared.ts  (reward block, ~L332–375)
+const gauge = VeloCLGauge.bind(gaugeAddress)
+const earnedResult = gauge.try_earned(nftOwner, tokenId)   // earned(address account, uint256 tokenId)
+
+if (!earnedResult.reverted) {
+  rewardAmount = earnedResult.value.toBigDecimal().div(BigDecimal.fromString("1e18"))  // AERO is 18-dec
+  const aeroPrice = getTokenPriceUSD(AERO, block.timestamp, false)                     // AERO/USDC pool
+  rewardUSD = rewardAmount.times(aeroPrice)
+} else {
+  log.warning(...)  // self-reporting — see §4
+}
+
+position.claimableReward       = rewardAmount
+position.claimableRewardUSD    = rewardUSD
+position.usdCurrent            = usd                 // base liquidity only
+position.usdCurrentWithRewards = usd.plus(rewardUSD) // base + rewards
+```
+
+`gaugeAddress` comes from `position.rewardsContract` if cached, else from the CL pool's
+`gauge()` getter (`pool.try_gauge()`), which is then cached on the position.
+
+### Why it matters
+
+`usdCurrentWithRewards` feeds `refreshPortfolio` → the portfolio snapshots → the **APR/ROI
+KPIs** surfaced on the website:
+
+- `DailyPopulationMetric.sma7dAPR` — "APR relative to USDC – MA7D"
+- `DailyPopulationMetric.sma7dEthAdjustedAPR` — "APR relative to ETH – MA7D"
+- `AgentPortfolioSnapshot.roi` — explorer per-day heatmap
+
+So a wrong reward value isn't cosmetic — it understates realized agent performance.
+
+---
+
+## 2. What is already verified (offline)
+
+- **`earned` ABI shape.** `abis/defi/VeloCLGauge.json` declares
+  `earned(address account, uint256 tokenId) → uint256`. The bound call matches the ABI.
+- **Gauge getters exist.** The same ABI exposes `rewardToken()`, `stakedContains(depositor,
+  tokenId)`, `stakedByIndex`, `stakedValues(depositor)` and `getReward` — the getters the
+  verification in §3 relies on.
+- **AERO token.** 18 decimals (so the `/1e18` is correct), address
+  `0x940181a94A35A4569E4529A3CDfB74e38FD98631`.
+- **AERO pricing.** Priced off the Aerodrome **AERO/USDC volatile** pool
+  `0x6cdcb1c4…` via the `velodrome_v2` adapter (`src/tokenConfig.ts`), Divya-confirmed.
+- **Contracts deployed.** Slipstream NFPM / factory and the CL pool `gauge()` getter all
+  return live bytecode on Base.
+
+## 3. What is NOT yet verified (needs a live staked position)
+
+Three assumptions can only be confirmed against real gauge state:
+
+1. **The gauge selector actually resolves.** Aerodrome's *deployed* CL gauge must implement
+   `earned(address,uint256)` with the same selector. If it differs, `try_earned` **reverts
+   silently** and reward records as **0** — no crash, no error. This is the dangerous mode:
+   the data looks healthy while undercounting.
+2. **The `account` argument is right.** We pass the **service safe** (`nftOwner`) as
+   `account`. Aerodrome credits `earned` to whoever **staked the NFT** (called
+   `gauge.deposit(tokenId)`). If Basius stakes directly from the safe → correct. If it stakes
+   through a relayer/intermediary → `earned(safe, tokenId)` returns 0 for the safe.
+3. **The full wei → AERO → USD product.** Each factor is individually checked; the
+   end-to-end number (`earned` wei → human AERO → USD → `usdCurrentWithRewards`) has never
+   been compared against a real, non-zero value.
+
+---
+
+## 4. Instrumentation added in this PR
+
+To convert "deferred unknown" into "self-reporting", the `else` branch of the `earned` call
+now emits a `log.warning` whenever `earned()` reverts on an **active** CL position
+(`src/veloCLShared.ts`). When the first real Basius position is staked, the indexer logs will
+immediately show whether the path works:
+
+- **No warning + non-zero `claimableRewardUSD`** → path is working; close this out.
+- **Warning fires** → assumption (1) or (2) is wrong; follow §5 to fix.
+
+No behavior change for the (correct) happy path; this only makes the failure visible.
+
+---
+
+## 5. Verification runbook (run once a Basius CL position is staked)
+
+Prereq: identify a staked position — a Basius service safe (607/610/611/612) that has called
+`gauge.deposit(tokenId)` on an Aerodrome Slipstream gauge. Use a **Base archive RPC**
+(`base.drpc.org`) so historical `eth_call` works.
+
+1. **Find the gauge.** From the position's pool: `eth_call pool.gauge()` → `gaugeAddress`.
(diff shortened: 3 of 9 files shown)


DIvyaNautiyal07:
Quick update on the verification status. One of Iason's Basius test agents (`0xcb9a1cb2b645f5fd69631abac212651b8ee0fc45`) staked into an Aerodrome **V2 stable** gauge during testing — pool `0x7A034374C89C463DD65D8C9BCfe63BcBCED41f4F` (USDC/eUSD), gauge `0x793f22ab88dc91793e5ce6adbd7e733b0bd4733e`. Verified on-chain at block 47702491.

This settles most of the "NOT verifiable until a position is staked" list:

- ✅ The service safe IS the actual depositor — the gauge `Deposit` log carries the safe as both indexed args, no operator/proxy in between. Applies universally (V2 + CL).
- ✅ `gauge.earned(safe)` resolves and returns non-zero — currently `0.0098155476 AERO` live on this position.
- ✅ `gauge.rewardToken() = 0x940181a9…` confirmed AERO.
- ✅ The deposited LP amount on-chain (`7157236262142` wei) matches our internal `current_liquidity` record exactly, so the agent's view of the position lines up with what the gauge sees.

Strict remaining gap: this stake hit the **V2 gauge**, which uses `earned(address)`. The CL path you implement reads `earned(address, uint256)`, which is a different selector on a different gauge ABI. The structural assumptions (depositor identity, rewardToken, account passed) all hold, so it's strongly suggested CL works too — but we can't claim CL is fully verified until a Basius position lands in a Slipstream pool. The `log.warning` guard you added is the right move for surfacing the gap if it ever fails on first CL stake.

DIvyaNautiyal07 (dismissed):
LGTM. Small surgical code change plus a well-scoped planning doc. Findings below are quality polish, not blockers.

**What's right**

- `log` is already imported in `veloCLShared.ts:1`, so the new `log.warning` doesn't need an import change.
- The warning fires only when `try_earned()` reverts on an active position; happy-path behavior is unchanged.
- Plan §2 offline-verified items are all accurate, including the AERO/USDC pool `0x6cdcb1c4…` confirmed in #156.
- The CL `earned(address,uint256)` claim is structurally correct: different selector from V2's `earned(address)`, which is the gap noted in [my comment](https://github.com/valory-xyz/autonolas-subgraph-studio/pull/159#issuecomment-4787928600) above.

**Nits (non-blocking)**

1. Sibling-handler asymmetry. The new branch logs, but earlier in the same function `pool.try_gauge()` reverting is a silent `return`:
   ```ts
   } else {
     position.usdCurrent = usd
     position.save()
     return
   }
   ```
   The asymmetry is defensible (no-gauge is a legitimate pool state, while gauge-earned-revert on an active stake is unexpected), but a one-line comment in the silent branch saying "pool has no gauge, legitimate state, do not log" would make the contrast intentional rather than accidental.

2. Plan §3 item (2) is already partially settled by [the on-chain check](https://github.com/valory-xyz/autonolas-subgraph-studio/pull/159#issuecomment-4787928600) on Iason's V2 stake: the service safe is the actual depositor (gauge `Deposit` emitted the safe as both indexed args). That's a universal claim across V2 and CL because Basius's staking flow doesn't change between them. The plan could call this out and narrow the still-unverified list to just §3 item (1), the CL-specific selector.

3. §6 symptom→fix table is missing one row. What if `gauge.earned()` returns a sane non-zero number but `claimableRewardUSD` indexes as 0? That points at the AERO price source returning 0 (priceDiscovery / pool reserves snapshot mismatch), not the gauge ABI. The current table jumps from "earned reverts" to "wrong by orders of magnitude" but doesn't cover "earned correct, USD product = 0."

4. Non-AERO reward token edge case isn't covered. Code hardcodes `/1e18` + `getTokenPriceUSD(AERO, …)`. If a future CL gauge emits a non-AERO reward, both assumptions break together. Aerodrome has standardized on AERO so this is low risk, but one sentence in §6 saying "if `rewardToken() != AERO`, decimals + pricing path needs rework, not just an ABI regen" would make the §5 step 2 check actionable.

5. No test exercises the new `log.warning` branch. Matchstick can't assert on logs cleanly and the whole point of the guard is to surface the gap via indexer logs at runtime, so this is fine. If you want belt-and-braces, a unit test that mocks `gauge.earned()` to revert and asserts the position still indexes cleanly (claimableRewardUSD == 0, no crash) would lock in the behavior.

kupermind:
Thanks @DIvyaNautiyal07 — this was exactly the verification needed, and your V2 example turned out to be more than a CL check.

**Acted on it:**
- Your confirmations are folded into `AERO-REWARDS-PLAN.md` §2: depositor == safe (universal V2+CL), `rewardToken` = AERO, `earned(safe)` non-zero, LP matches. The **only strict gap left is the CL `earned(address,uint256)` selector** (your stake hit the V2 `earned(address)`), and the `log.warning` guard surfaces it on the first live CL stake.
- Your V2 stake also exposed a real bug: because we valued V2 positions from `pool.balanceOf(safe)`, staking the LP into the gauge (LP leaves the safe → `balanceOf == 0`) made the position **vanish** from the portfolio. **Fixed in this PR** — `refreshVeloV2Position` now resolves the gauge via `Voter.gauges(pool)`, counts `gauge.balanceOf(safe)` as staked LP, and folds `gauge.earned(safe)` into `usdCurrentWithRewards`. I re-verified the whole chain against your position (gauge `0x793f…`, block 47702491): `balanceOf(safe)` = `7157236262142`, matching your record. Tests added.

One follow-up question: this gap is inherited verbatim from **babydegen-optimism** — do Optimus agents stake V2 LP into gauges there? If so I'll mirror the fix in a separate PR. (If Optimus is CL-only, optimism is unaffected in practice.)

DIvyaNautiyal07 (dismissed):
Answer to the follow-up: **yes, Optimus stakes V2 LP into Velodrome gauges on Optimism.** `evaluate_strategy.py:3322` (`_build_stake_lp_tokens_action`) has explicit `if is_cl_pool / else` branches, and both produce a `STAKE_LP_TOKENS` action that dispatches to `pools/velodrome.py:2774` (`stake_lp_tokens`), which is chain-parameterised. So babydegen-optimism has the same KPI bug today (staked V2 LP vanishes from the portfolio because `pool.balanceOf(safe) == 0` after staking). Worth mirroring this fix in a separate PR for optimism.

On the new commit (`728d71a3`):

- `resolveV2Gauge` is right not to cache `Address.zero()`. Pools that get gauged later are picked up on a subsequent refresh.
- `userBalance = userBalanceInSafe.plus(stakedBalance)` is correct because Aerodrome v2 gauges hold the LP token (staking is a transfer, not a burn), so `pool.totalSupply` already includes the staked LP. The inline comment captures that exactly.
- Reward path mirrors the CL path: `earned()` → `/1e18` → AERO/USDC pricing → `usdCurrentWithRewards`.
- Zero-LP close-out now zeroes `usdCurrentWithRewards`, `claimableReward`, `claimableRewardUSD` too. Good.
- Tests cover both branches (staked-only with reverting `earned`, and staked with non-zero `earned`). Pinning the test constants to gauge `0x793f22ab…` is nice because it doubles as an integration anchor to the live position we verified.

One small nit, non-blocking: the new V2 path doesn't add a `log.warning` on `gauge.try_earned()` revert, even though that's the same silent-zero failure mode the CL log.warning was designed to surface. If a V2 gauge ever ships with a non-standard `earned` ABI, V2 rewards quietly record 0 with no signal. Symmetric treatment with the CL path (one-line `log.warning` in the `try_earned` else-branch on V2) would catch it. Same applies to `try_balanceOf`, though that's much less likely on a deployed ERC20-like gauge, so the `earned` warning is the higher-value one.

One thing worth flagging explicitly: unlike the CL change (instrumentation only), this V2 fix is a real KPI change. V2 staked positions now contribute to `usdCurrentWithRewards`, which feeds the website APR/ROI. Worth a callout in the release notes when this ships.

kupermind:
Thanks for the thorough review @DIvyaNautiyal07 — and for confirming the Optimism side. Addressed the nits in `0363dbb`:

- **V2 `earned` silent-zero guard (your approved-review nit):** added a symmetric `log.warning` in the V2 `try_earned()` else-branch, matching the CL guard. `try_balanceOf` left unguarded as you suggested (much lower risk on a deployed gauge).
- **CL silent-return comment (nit 1):** the no-gauge `return` branch now carries a comment explaining it's a legitimate pool state and deliberately not logged — so the asymmetry with the `earned`-revert warning is intentional, not accidental.
- **Plan §6 table (nits 3 & 4):** added a row for "`earned` correct but `claimableRewardUSD` = 0" → AERO **price source**, not the gauge ABI; and a row for "`rewardToken() != AERO`" → decimals + pricing both break, not just an ABI regen.
- **Plan §3 item (2) (nit 2):** already narrowed to the CL selector in the doc — item (2) is struck through as settled by your on-chain depositor check.
- **KPI-change callout:** added a release-notes note in §8 — V2 staked positions now contribute value + AERO rewards to `usdCurrentWithRewards` (previously $0 while staked), which feeds website APR/ROI. Will flag at ship time.
- Nit 5 (log-branch test): the new V2 staked test already asserts the position indexes cleanly with `earned` reverting (`claimableRewardUSD == 0`, stays active), which locks in the no-crash behavior; agreed Matchstick can't assert the log line itself.

**Optimism follow-up:** confirmed needed — thanks for the `evaluate_strategy.py` / `velodrome.py` pointers. I'll mirror the V2 staked-position fix to `babydegen-optimism` in a separate PR.

DIvyaNautiyal07 (approved):
Verified the two new commits (`0363dbb0` and `0510ff78`).

`0363dbb0` lands the doc and CL-comment items I raised on this PR specifically: AERO-REWARDS-PLAN.md §6 has the two new symptom rows (price-source-zero, non-AERO rewardToken), §7 confirms the Optimus V2 staking gap and adds the release-notes callout, and `veloCLShared.ts` now has an inline comment in the no-gauge branch explaining the asymmetry with the `earned()`-revert log.warning. The symmetric V2 `earned()` log.warning is in place too.

`0510ff78` mirrors PR #160's fix commit byte-for-byte (with Base addresses swapped in): `resolveV2Gauge` splits the Voter-revert case from the legitimate no-gauge case, `gauge.try_balanceOf()` revert has its own log.warning, both V2 warnings lead with the pool address, the AERO/USDC mock has the correct on-chain token ordering (USDC `0x8335…` < AERO `0x9401…` so USDC is token0), and the three new tests (no-gauge close-out, cached-gauge fast path, exited-staked-position clear-out) match #160's coverage.

base and optimism are now structurally consistent. LGTM.

Tanya-atatakai (approved):
Reviewed in an isolated worktree — read the full diff, traced the runtime trigger path, and ran `codegen` + `build` + `test` (24/24 green). Approving.

**The fix is correct where it's easy to get wrong:**
- Share math is sound — `pool.totalSupply` already includes staked LP (staking is a transfer, not a burn), so adding `gauge.balanceOf` to the numerator keeps `userShare = userBalance/totalSupply` correct. No double-correction.
- No reward double-counting — `usdCurrentWithRewards = usdCurrent + rewardUSD` is recomputed (assignment, not accumulation); AERO is a separate token not yet in LP reserves.
- The inactive branch now zeros `usdCurrentWithRewards`/`claimableReward`/`claimableRewardUSD` (previously these would go stale on close) — nice catch.
- Gauge-cache retry semantics are right: Voter-revert and zero-gauge are not cached (so a later-gauged pool is picked up), a resolved gauge is cached and `pp.save()` persists it.
- End-to-end freshness confirmed: the daily UTC-midnight block handler re-refreshes every *active* position, so a staked position stays both active **and** re-valued daily — not just correct at the staking event.

CL `log.warning` guard + `AERO-REWARDS-PLAN.md` tracking of the unverified `earned(address,uint256)` selector is the right call.

**Non-blocking observations (follow-ups, not merge blockers):**
1. **Stale gauge cache** — once `rewardsContract` is cached it's never re-resolved. If Aerodrome kills a gauge and deploys a new one for the same pool, the position keeps reading the dead gauge (`earned` reverts → logged; `balanceOf` 0 → a still-staked position could be misread as closed). Rare, but worth a comment/follow-up.
2. **`gauge.balanceOf` revert edge** — if `balanceOf` reverts while the LP is genuinely staked (safe balance 0), `userBalance` becomes 0 and the position flips inactive: the exact symptom being fixed. Standard ERC20 `balanceOf` won't revert so risk is low and it's logged, but it could be hardened by not closing a position when a gauge exists but `balanceOf` reverted.

Note for whoever merges second of #159/#160: the two new ABI files are byte-identical and new vs `main`, so it's an identical re-add — no conflict. Also a **KPI callout** for release notes: staked V2 positions now contribute principal + rewards to `usdCurrentWithRewards` (previously $0 while staked), which feeds APR/ROI.

PR #160, opened 2026-06-24
fix(babydegen-optimism): index VELO gauge rewards + value staked V2 positions

kupermind:
## What & why

Mirror of #159 for **Optimism**. @DIvyaNautiyal07 confirmed Optimus **stakes V2 LP into Velodrome gauges** on Optimism (`evaluate_strategy.py` `_build_stake_lp_tokens_action` → `velodrome.py` `stake_lp_tokens`, chain-parameterised), so `babydegen-optimism` has the **same KPI bug** #159 fixed on Base:

V2 position value is read from `pool.balanceOf(safe)`. Staking LP into a gauge **moves the LP out of the safe** → `balanceOf == 0` → the position was marked `isActive=false` and zeroed. A staked, earning V2 position **vanished** from the portfolio (and its VELO rewards were never counted — the old `// TODO: Calculate V2 fees later`).

## The fix (`refreshVeloV2Position`)

- **Resolve the gauge** via the Velodrome **Voter** — `Voter.gauges(pool)` (`VELO_VOTER 0x41C914ee…`), cached on `rewardsContract`.
- **Count staked LP** — `effectiveBalance = pool.balanceOf(safe) + gauge.balanceOf(safe)`. `pool.totalSupply` already includes staked LP (staking is a transfer, not a burn), so the share math is unchanged. A staked-only position is no longer misread as closed.
- **Add rewards** — `gauge.earned(safe)` (VELO, 18-dec, priced via `priceDiscovery`) → `claimableReward/USD` and `usdCurrentWithRewards`.
- **Silent-zero guard** — symmetric `log.warning` on `earned()` revert, matching the CL path.

New ABIs (shared, identical to #159): `abis/defi/VeloVoter.json`, `abis/defi/VeloV2Gauge.json`, wired into the `VeloV2Pool` template.

## Verified on-chain (Optimism)
Sampled a real gauged Velodrome V2 pool: `Voter.gauges(pool)` → gauge ✅ · `gauge.stakingToken()` == pool ✅ · `gauge.rewardToken()` = VELO (`0x9560…`) ✅ · `gauge.earned(address)` + `gauge.balanceOf(address)` selectors resolve ✅. The Velodrome V2 gauge ABI is identical to Aerodrome's (fork), so the same ABI files are reused.

## Tests
`yarn codegen` + `build` + `test` green — **12/12** (＋2): a staked position (`pool.balanceOf==0`, gauge holds the LP) stays active with `liquidity` = staked amount; `gauge.earned` folds into `usdCurrentWithRewards`.

## Notes
- **KPI change** (same as #159): V2 staked positions now contribute value + VELO rewards to `usdCurrentWithRewards`, which feeds APR/ROI — previously $0 while staked. Worth a release-notes callout when this ships.
- The two new ABI files are byte-identical to the ones in #159; whichever merges second is a trivial identical-add (no conflict).

### DIFF
--- a/abis/defi/VeloV2Gauge.json
+++ b/abis/defi/VeloV2Gauge.json
@@ -1,0 +1,30 @@
+[
+  {
+    "type": "function",
+    "name": "balanceOf",
+    "stateMutability": "view",
+    "inputs": [{ "name": "account", "type": "address" }],
+    "outputs": [{ "name": "", "type": "uint256" }]
+  },
+  {
+    "type": "function",
+    "name": "earned",
+    "stateMutability": "view",
+    "inputs": [{ "name": "account", "type": "address" }],
+    "outputs": [{ "name": "", "type": "uint256" }]
+  },
+  {
+    "type": "function",
+    "name": "rewardToken",
+    "stateMutability": "view",
+    "inputs": [],
+    "outputs": [{ "name": "", "type": "address" }]
+  },
+  {
+    "type": "function",
+    "name": "stakingToken",
+    "stateMutability": "view",
+    "inputs": [],
+    "outputs": [{ "name": "", "type": "address" }]
+  }
+]
--- a/abis/defi/VeloVoter.json
+++ b/abis/defi/VeloVoter.json
@@ -1,0 +1,9 @@
+[
+  {
+    "type": "function",
+    "name": "gauges",
+    "stateMutability": "view",
+    "inputs": [{ "name": "pool", "type": "address" }],
+    "outputs": [{ "name": "", "type": "address" }]
+  }
+]
--- a/subgraphs/babydegen/babydegen-base/AERO-REWARDS-PLAN.md
+++ b/subgraphs/babydegen/babydegen-base/AERO-REWARDS-PLAN.md
@@ -1,0 +1,227 @@
+# AERO gauge-reward indexing — verification & implementation plan
+
+**Status:** CL reward path is **built and shipped** (in `#156`). Most assumptions are now
+**verified on-chain** via a real Basius **V2** stake (Divya, block 47702491 — §2): depositor
+identity, `rewardToken = AERO`, and `earned(safe)` resolving non-zero all hold. The **one
+strictly-open gap is the CL `earned(address,uint256)` selector** (a different gauge ABI than
+the V2 `earned(address)` that was exercised) — unverifiable until a Basius position lands in a
+Slipstream pool. Divya's example also surfaced a **separate, larger V2 gap** (§7) — staked V2
+positions vanished from the portfolio — which **this PR fixes** (gauge-aware V2 valuation +
+rewards, with tests). This doc captures the path, the verification status, and the runbook to
+close CL out once a real staked CL position exists.
+
+This is deliberately a docs-first PR: there is **no Basius gauge state on-chain to test
+against today**, so we describe the plan and add self-reporting instrumentation rather than
+guess at fixes.
+
+---
+
+## 1. What the path does today
+
+For every **active Aerodrome Slipstream (CL)** position, `refreshVeloCLPosition`
+(`src/veloCLShared.ts`) reads unclaimed staking rewards from the position's gauge and folds
+them into the portfolio value:
+
+```ts
+// src/veloCLShared.ts  (reward block, ~L332–375)
+const gauge = VeloCLGauge.bind(gaugeAddress)
+const earnedResult = gauge.try_earned(nftOwner, tokenId)   // earned(address account, uint256 tokenId)
+
+if (!earnedResult.reverted) {
+  rewardAmount = earnedResult.value.toBigDecimal().div(BigDecimal.fromString("1e18"))  // AERO is 18-dec
+  const aeroPrice = getTokenPriceUSD(AERO, block.timestamp, false)                     // AERO/USDC pool
+  rewardUSD = rewardAmount.times(aeroPrice)
+} else {
+  log.warning(...)  // self-reporting — see §4
+}
+
+position.claimableReward       = rewardAmount
+position.claimableRewardUSD    = rewardUSD
+position.usdCurrent            = usd                 // base liquidity only
+position.usdCurrentWithRewards = usd.plus(rewardUSD) // base + rewards
+```
+
+`gaugeAddress` comes from `position.rewardsContract` if cached, else from the CL pool's
+`gauge()` getter (`pool.try_gauge()`), which is then cached on the position.
+
+### Why it matters
+
+`usdCurrentWithRewards` feeds `refreshPortfolio` → the portfolio snapshots → the **APR/ROI
+KPIs** surfaced on the website:
+
+- `DailyPopulationMetric.sma7dAPR` — "APR relative to USDC – MA7D"
+- `DailyPopulationMetric.sma7dEthAdjustedAPR` — "APR relative to ETH – MA7D"
+- `AgentPortfolioSnapshot.roi` — explorer per-day heatmap
+
+So a wrong reward value isn't cosmetic — it understates realized agent performance.
+
+---
+
+## 2. What is already verified
+
+### Offline (ABI / token / contracts)
+
+- **`earned` ABI shape.** `abis/defi/VeloCLGauge.json` declares
+  `earned(address account, uint256 tokenId) → uint256`. The bound call matches the ABI.
+- **Gauge getters exist.** The same ABI exposes `rewardToken()`, `stakedContains(depositor,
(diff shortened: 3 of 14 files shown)


DIvyaNautiyal07:
LGTM overall, mirror of #159 is clean. Verified `VELO_VOTER 0x41C914ee…` against Optimus `liquidity_trader_abci/skill.yaml:336` (Optimism velodrome voter) and the VELO/USDC price pool `0xa0A215dE…` against `babydegen-optimism/tokenConfig.ts` VELO entry. ABI files, gauge resolution math (`pool.balanceOf + gauge.balanceOf` with unchanged `totalSupply`), and the new `log.warning` symmetric with the CL path are all correct.

A few things worth tightening before merge. None are correctness-blocking against real data (we already verified the V2 fix works on Iason's Base position in #159), but several are silent-failure modes the rest of this PR was explicitly designed to surface.

**P1**

1. `tests/mapping.test.ts` ~L497 — token0/token1 inversion in `mockVeloUsdcPool`. The mock sets `token0=VELO, token1=USDC`, but on-chain `USDC 0x0b2c… < VELO 0x9560…`, so production hits the `token1` branch in `getVelodromeV2Price` (`priceAdapters.ts:288`), not the `token0` branch the test exercises. Math is symmetric so both branches return ~1.3 and the test passes, but a future regression in the token1 path would silently pass too. Fix: swap token0 and token1 in the mock and swap the reserve values.

2. `src/veloV2Shared.ts` ~L376 — silent-zero on `gauge.try_balanceOf` revert. Mirrors the exact failure mode `try_earned` got a `log.warning` for, but here `balanceOf` reverting silently records `stakedBalance=0`, which combined with `userBalanceInSafe=0` (the normal staked state) flips `isActive=false` and zeros the position. Add a symmetric `log.warning` on the `else` branch.

3. `src/veloV2Shared.ts` ~L25 — silent-zero on `voter.try_gauges` revert. `resolveV2Gauge` collapses "Voter reverted" and "no gauge for pool" into the same `Address.zero()` return. The Voter-reverted case is a transient failure that would silently undervalue a staked position. Suggest splitting the branches and logging on `res.reverted` specifically.

4. `tests/mapping.test.ts` — no test for `Voter.gauges()` returning `0x0`. That's the default state for any pool without a gauge, and production code's bind-on-zero scenario goes untested.

5. `tests/mapping.test.ts` — no test for the cached-gauge fast path. `makeOpenV2Position` never sets `pp.rewardsContract`, so every test run goes through the Voter lookup. A regression casting `Bytes` back to `Address` in the cached branch would not be caught.

6. `tests/mapping.test.ts` ~L496 — `mockVeloUsdcPool` hardcodes `0xa0A215dE…`. If `tokenConfig.ts` ever picks a different VELO/USDC price source, the test silently keeps mocking a stale pool while production queries a new one, and the assertions still pass. Read the pool address from a shared constant.

**P2**

7. `resolveV2Gauge` is called from exactly one site. The cache-or-fetch pattern doesn't earn its own function until a second caller exists, optional simplification.

8. Inactive-path zeroing of `claimableReward` / `claimableRewardUSD` / `usdCurrentWithRewards` is new behavior with no dedicated test. The case where both `pool.balanceOf=0` AND `gauge.balanceOf=0` (genuinely exited staked position) is not exercised. A regression that left stale reward fields on close would pass.

9. `usdCurrent + rewardUSD` never tested with non-zero `usdCurrent`. Both tests use dummy tokens so `usdCurrent=0`. If the `+` got changed to `=`, test 2 still passes (0 + 2.6 == 2.6).

**Nit**

10. `positionId.toHexString()` in the new `log.warning` is a long opaque hex blob, awkward to grep. Consider leading with the pool address or a more readable handle.

11. `createService`, `makeOpenV2Position`, `mockV2PoolState`, `v2Block` are duplicated verbatim from PR #159's babydegen-base tests. No shared cross-subgraph test infra exists today, so this is a structural drift gap rather than a single-PR fix, but worth noting since the pattern will recur.

kupermind:
Thanks @DIvyaNautiyal07 — addressed the silent-failure items (they're squarely in the spirit of this PR). Pushed to #160, and **mirrored the same changes to #159** so base/optimism stay identical.

**Done (code):**
- **#3** `resolveV2Gauge` now splits the **Voter-revert** case (transient → `log.warning`, not cached so it retries) from the legitimate **no-gauge** case (silent, not cached so a later-gauged pool is still picked up).
- **#2** `log.warning` on `gauge.balanceOf()` revert — same silent-zero mode as `earned()` (a revert there would flip a staked position to closed).
- **#10** both V2 `log.warning`s now lead with the **pool address** instead of the opaque `positionId` hex blob.

**Done (tests):**
- **#1** Fixed the `token0/token1` ordering in the price mock to match on-chain (USDC < VELO on OP, USDC `0x8335…` < AERO `0x9401…` on Base — verified), so the test now exercises the same `getVelodromeV2Price` branch production hits.
- **#4** no-gauge (`Voter.gauges` → `0x0`) → position closes cleanly.
- **#5** cached-gauge fast path — `rewardsContract` pre-set and `Voter.gauges` intentionally **not** mocked, so if the cache were ignored matchstick would throw.
- **#8** genuinely-exited staked position (pool & gauge balance both 0) → `claimableReward`/`claimableRewardUSD`/`usdCurrentWithRewards` cleared (pre-seeded stale values prove it).

OP: 15 tests; Base: 24 tests. All green.

**Deferred, with rationale:**
- **#6** (read pool from a shared constant): pulled the pool address into a single test-file `const` with a comment tying it to `tokenConfig.ts`. A truly shared constant means exporting the price-source addresses out of `tokenConfig`'s runtime map — a structural change I'd rather not fold into this fix.
- **#7** (inline `resolveV2Gauge`): kept as a function — the cache-or-fetch + dual-logging is more readable named than inlined, and #160 already gives it a second use site shape.
- **#9** (`+` vs `=` with non-zero `usdCurrent`): the dummy-token setup keeps `usdCurrent=0`; forcing it non-zero pulls the full token-pricing pipeline into the test. Noting it as a known coverage gap rather than adding brittle pricing mocks — open to a follow-up if you feel strongly.
- **#11** (cross-subgraph shared test infra): real, but structural — there's no shared test harness across subgraphs today, so I kept the helpers duplicated (now identical in both PRs). Worth a separate infra ticket.

DIvyaNautiyal07 (approved):
Verified the fix commit (`bf5f4b4f`). All addressable items from the review landed:

- #1 token0/token1 ordering in the price mock now matches on-chain (USDC is token0). The test exercises the same `getVelodromeV2Price` branch production hits.
- #2 / #3 the two silent-zero modes (`gauge.try_balanceOf` and `voter.try_gauges` reverts) both have `log.warning` now. `resolveV2Gauge` cleanly separates the transient-revert case from the legitimate "no gauge" case.
- #4 no-gauge test covers `Voter.gauges → 0x0`.
- #5 cached-gauge test pre-seeds `pp.rewardsContract` and intentionally omits the `Voter.gauges` mock, so matchstick would throw if the cache were bypassed. Nice design.
- #6 pool address pulled into a named test-file constant with a comment tying it to `tokenConfig.ts`.
- #8 exited-staked-position test pre-seeds stale reward fields and asserts they clear.
- #10 both V2 warnings now lead with the pool address.

Deferrals (#7 inline, #9 non-zero `usdCurrent`, #11 shared test infra) are all reasonable with the rationale given.

Plus the same changes mirrored back to #159 so base/optimism stay identical. LGTM.

Tanya-atatakai (approved):
Reviewed in an isolated worktree alongside #159 — read the full diff and ran `codegen` + `build` + `test` (15/15 green). Approving.

Clean mirror of #159 for Optimism/Velodrome. Verified the parts that matter:
- Share math sound (`pool.totalSupply` includes staked LP → adding `gauge.balanceOf` is correct).
- No reward double-count (`usdCurrentWithRewards = usdCurrent + rewardUSD`, recomputed each refresh).
- Inactive branch zeros the reward fields.
- Gauge-cache retry semantics correct (revert/zero not cached; resolved gauge cached + persisted).
- `VELO` constant defined and priced via priceDiscovery; daily scheduler re-refreshes active positions so staked positions stay current.

The two new ABI files are byte-identical to #159's and new vs `main` — whichever merges second is an identical re-add, no conflict.

**Non-blocking (same as #159):**
1. Stale gauge cache — `rewardsContract` is never re-resolved once cached; a redeployed gauge for the same pool would be read stale.
2. `gauge.balanceOf` revert while genuinely staked would flip the position inactive (the symptom being fixed); standard ERC20 won't revert, and it's logged.

**KPI callout** for release notes: V2 staked positions now contribute value + VELO rewards to `usdCurrentWithRewards`, previously $0 while staked.

PR #164, opened 2026-07-08
fix(babydegen-base): stop re-saving immutable SwapTransaction

DIvyaNautiyal07:
## Summary

The redeployed `olas-babydegen-base` subgraph halted with:

```
Failed to transact block operations: internal error: immutable entity type SwapTransaction only allows inserts, not Overwrite ...
```

`SwapTransaction` is declared `@entity(immutable: true)` in `schema.graphql`. Two code paths loaded it and called `save()` on it, which the Graph node rejects. Before #163 the offending path was gated behind the dead entry-amount fallback and never actually ran; the fallback fix woke it up and every backfilled v2 position with a swap in its association window now halts the indexer.

## Fix

- **`helpers.ts:associateSwapsWithPosition`**: dropped the load-and-save loop that set `swap.isAssociated = true`. `isAssociated` has no readers anywhere in the code (schema field, two writers, zero reads), so removing the write is behaviourally a no-op. The slippage accounting that feeds `ProtocolPosition.swapSlippageUSD` runs above this loop and is unchanged.
- **`swapTracking.ts:associateSwapWithPosition`**: removed the whole helper. It had no callers and would hit the same overwrite the moment it was wired up.
- Cleaned up the now-unused imports (`SwapTransaction` in `helpers.ts`, `SwapToEntryAssociation` in `swapTracking.ts`).

## Test

New matchstick regression (`backfill fallback with a matching buffered swap does not overwrite the immutable SwapTransaction`) seeds:
- a `ProtocolPosition` with `entryAmountUSD = 0` (the state that triggers the fallback)
- a stored `SwapTransaction` with `isAssociated = false`
- an `AgentSwapBuffer` bucket entry referencing the swap, timestamped inside the association window

then calls `refreshVeloV2Position`. Before this change matchstick would trip the immutability check; now the call completes, the position's `swapSlippageUSD` equals the buffered slippage (0.01197), and the stored `SwapTransaction` is untouched.

All 30 base tests pass locally (was 29). Ran `yarn audit:prod`, `yarn audit:install-hooks`, and lockfile-lint on the base path per `.github/workflows/ci.yml`.

## Deployment

Redeploy only, no resync. The buggy state on the live deployment is a halted indexer, not corrupted data.

## Follow-ups

- The same immutable-overwrite exists in `babydegen-optimism` (identical `helpers.ts`, same schema). It is latent there because the same entry-amount fallback is still gated by the dead `entryTimestamp == 0` condition. Wanted to keep this PR to the acute Basius fix; a separate PR should port both changes to optimism.

### DIFF
--- a/.supply-chain/audit-allowlist.json
+++ b/.supply-chain/audit-allowlist.json
@@ -9,5 +9,15 @@
     "added": "YYYY-MM-DD — REQUIRED",
     "review": "YYYY-MM-DD — REQUIRED, expired prints warning but does not fail"
   },
-  "entries": []
+  "entries": [
+    {
+      "id": 1122670,
+      "ghsa": "GHSA-mp2f-45pm-3cg9",
+      "package": "decompress",
+      "severity": "critical",
+      "reason": "Build-time only via @graphprotocol/graph-cli. Path traversal on archive extraction; graph-cli extracts vendor-published npm tarballs, not untrusted input. No upstream fix (fix in <0.0.0), no released graph-cli drops decompress. Revisit if graph-cli publishes a version without decompress or if a fixed decompress ships.",
+      "added": "2026-07-08",
+      "review": "2026-10-08"
+    }
+  ]
 }
--- a/subgraphs/babydegen/babydegen-base/src/helpers.ts
+++ b/subgraphs/babydegen/babydegen-base/src/helpers.ts
@@ -1,12 +1,11 @@
 import { BigDecimal, BigInt, Address, Bytes, log, ethereum } from "@graphprotocol/graph-ts"
-import { 
-  FundingBalance, 
-  AgentPortfolio, 
+import {
+  FundingBalance,
+  AgentPortfolio,
   AgentPortfolioSnapshot,
   ProtocolPosition,
   Service,
-  AgentSwapBuffer,
-  SwapTransaction
+  AgentSwapBuffer
 } from "../generated/schema"
 import { calculateUninvestedValue, updateFundingBalance } from "./tokenBalances"
 import { getServiceByAgent } from "./config"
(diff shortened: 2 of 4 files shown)


Tanya-atatakai (dismissed):

DIvyaNautiyal07:
Heads up on the failing "Dependency audit (root tree)" check, it's not from the code in this PR.

GitHub just promoted GHSA-mp2f-45pm-3cg9 (path traversal in `decompress`) to Critical. Nothing in the repo imports decompress directly, it sits transitively under `@graphprotocol/graph-cli` and only runs at build time when graph-cli unpacks tarballs it fetched from npm and The Graph registry. No upstream fix exists (the advisory records `fix in <0.0.0`, and `@graphprotocol/graph-cli@latest` is still 0.98.1, which is what we pin). Every open PR after this promotion will hit the same gate.

The audit script prints three options:

- Bump the dep. Not possible right now, no fixed version of decompress exists.
- Add a Yarn `resolutions:` override. Same problem, nothing to point at.
- Add an entry to `.supply-chain/audit-allowlist.json`. This is the documented "no fix available" escape hatch in SUPPLY-CHAIN-SECURITY.md.

The allowlist is currently empty, so this would be its first entry and I don't want to write it unilaterally. Practical risk in our setup is low (build-time only, graph-cli only extracts vendor-published tarballs, none of this reaches the deployed subgraph WASM), but it is a shared-config precedent.

Proposed entry:

```json
{
  "id": 1122670,
  "ghsa": "GHSA-mp2f-45pm-3cg9",
  "package": "decompress",
  "severity": "critical",
  "reason": "Build-time only via @graphprotocol/graph-cli. Path traversal on archive extraction; graph-cli extracts vendor-published npm tarballs, not untrusted input. No upstream fix (fix in <0.0.0), no released graph-cli drops decompress. Revisit if graph-cli publishes a version without decompress or if a fixed decompress ships.",
  "added": "2026-07-08",
  "review": "2026-10-08"
}
```

Options for how to proceed:

1. Add the allowlist entry on this branch so #164 can merge. Unblocks the babydegen-base indexer fix immediately, at the cost of mixing two concerns in one PR.
2. Add the allowlist entry on a separate branch and its own PR. Cleaner history, but delays #164 landing.
3. Different reason wording, different review date, or a different approach entirely.

Happy to go with whichever you prefer.

atepem (dismissed):

Tanya-atatakai (approved):

atepem (approved):

